Issue #5, opened 2016-12-13
Update package.json to reflect open source contribution


pmonks:
Since transfer to the open source repository, a couple of items need to be updated in [`package.json`](https://github.com/symphonyoss/SymphonyElectron/blob/master/package.json); specifically:

1. the git URL has changed to `https://github.com/symphonyoss/SymphonyElectron.git` ([line 28](https://github.com/symphonyoss/SymphonyElectron/blob/master/package.json#L28))
2. the license should be listed as `Apache License, Version 2.0` ([line 34](https://github.com/symphonyoss/SymphonyElectron/blob/master/package.json#L34))

PR #6, opened 2016-12-13
Fix issue #5 & #7, and add .gitignore

pmonks:
This PR fixes issue #5 and issue #7, and adds a .gitignore to avoid build droppings.  It also switches the default Symphony URL to the Foundation's [Open Developer Platform](https://symphonyoss.atlassian.net/wiki/display/FM/Open+Developer+Platform) (since it's more likely that open source developers will be able to gain access to that pod).


package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "Symphony",
-  "productName": "Symphony",
+  "productName": "Symphony - ODP",
```
lneir: what is ODP?  can we just keep as symphony?

package.json
```diff
@@ -1,8 +1,8 @@
...
   "version": "1.0.1",
-  "description": "Symphony desktop app",
+  "description": "Symphony desktop app (Foundation ODP)",
```
lneir: why ODP?

package.json
pmonks: The ODP is the Foundation's [Open Developer Platform](https://symphonyoss.atlassian.net/wiki/display/FM/Open+Developer+Platform).

package.json
pmonks: Because it's more likely that the typical contributor to or consumer of SymphonyElectron has access to the ODP than the Symphony corporate pod.

pmonks: Of course it would be even better if there was some way for SymphonyElectron to autodetect (or, less ideally, let the user choose) the pod they're connecting to.

lneir:
please squash down to single commit and we can merge.

pmonks:
It may make more sense to leave the commits separate, so that they can be individually correlated to issues.  What's the concern with the current set?

lneir:
i'll squash and merge using github.  gives cleaner history on master.

PR #39, opened 2017-03-23
Add context menu to Symphony Electron

rbarros-daitan:
Electron-38


js/main.js
```diff
@@ -47,6 +51,8 @@ function getUrlAndOpenMainWindow() {
         let title = 'Error loading configuration';
         electron.dialog.showErrorBox(title, title + ': ' + err);            
     });
+
+    applyContextMenu();
```
lneir: can we instead move this into windowMgr.js code and put in createMainWindow and createChildWindow and then pass window object in via opts parameter...e.g., applyContextMenu({ window: mainWindow});  and plyContextMenu({ window: childWindow});

js/contextMenu.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
```
lneir: also you could create a subdir for the code: js/menus and move both this and menuTemplate.js into that dir.

js/main.js
rbarros-daitan: Do we plan to have different right click menu based on child or main window?
Current implementation applies the same context menu on main window and on any window created after that.

js/contextMenu.js
rbarros-daitan: ok, done.

js/main.js
lneir: might be diff but for now the same.  also want to move code into here since this involves creating windows.

Issue #43, opened 2017-03-25
Broader gitignore support


VishwasShashidhar:
We should add a broader support for gitignore to ignore files generated by the following components

Node, VisualStudio, Eclipse, Jetbrains, Linux, Netbeans, SublimeText, TextMate, Vim, Windows, macOS, VisualStudioCode.

The updated file will be as attached.
[gitignore-updated.txt](https://github.com/symphonyoss/SymphonyElectron/files/869695/gitignore-updated.txt)

ivishwas-s:
I have a branch ready with this change, however, I don't have access to push it to the repo. I want to push the branch and create a PR. Can anyone help?

lneir:
we use fork and pull model here.

ivishwas-s:
Done. Please review and merge.

PR #44, opened 2017-03-28
added more rules to gitignore (#43) and fixed typo in readme

VishwasShashidhar:
Please review and merge.


.gitignore
```diff
@@ -1,4 +1,574 @@
 node_modules
-dist
 .DS_Store
 js/preload/_*.js
+
```
lneir: this is a just a super generic list and many don't apply to this project.  to prevent pollution this should be stripped heavily stripped down.   e.g., why mongoSettings.xml, .pyc, etc.   i would prefer to keep minimal and add as necessary.

ivishwas-s: ok, makes sense. I've stripped down a lot of the rules related to plugins/add-ons. Please see if the latest commit can be merged.

The idea behind adding a lot of these rules are to ensure that people using different tools don't accidentally commit things and then we end up rolling that back.

ivishwas-s: and, shouldn't the config file by a local copy?

lneir:
closing - if still needed please submit appropriate subset.

PR #47, opened 2017-03-29
Enable the Chromium sandbox

ColinEberhardt:
According to recently added documentation, `sandbox: true` creates a sandbox 'compatible' renderer process, but doesn't actually enable the sandbox. The command line argument I've added is also required.

See discussion here: https://github.com/electron/electron/pull/9054#discussion_r108695652

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "author": "Symphony",
   "main": "js/main.js",
   "scripts": {
-    "dev": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron .",
-    "demo": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron . --url=file://$(pwd)/demo/index.html",
+    "dev": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron . --enable-sandbox",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,8 +6,8 @@
   "author": "Symphony",
   "main": "js/main.js",
   "scripts": {
-    "dev": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron .",
-    "demo": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron . --url=file://$(pwd)/demo/index.html",
+    "dev": "npm run browserify-preload && cross-env ELECTRON_DEV=true electron . --enable-sandbox",
```
lneir: it should already be enabled...  see: when creating main window: https://github.com/symphonyoss/SymphonyElectron/blob/master/js/windowMgr.js#L48 and also for and child windows.

ColinEberhardt: You'd think so, but it doesn't. See the comment in the above discussion from the electron contributor who added sandbox support:

> Passing sandbox alone does not enable chromium os sandbox, it simply creates a renderer that is compatible with it.

ColinEberhardt: Note, the comment is on an out-dated diff which you might have to expand

lneir: good info, thanks.  above change will effect dev and demo.  also need to figure out how to add for builder.  maybe electron-builder has an option.  i'll look into.

PR #61, opened 2017-04-19
auto launch for windows.

rbarros-daitan:
Code implements ELECTRON-25 which describes AutoLaunch for Windows settings. Code also prepared the settings for MinimizeOnClose param.

### DIFF
--- a/config/Symphony.config
+++ b/config/Symphony.config
@@ -1,3 +1,5 @@
 {
     "url": "https://foundation-dev.symphony.com"
+    "launchOnStartup": true,
+    "minimizeOnClose": true
 }
--- a/js/getConfig.js
+++ b/js/getConfig.js
@@ -7,6 +7,7 @@ const fs = require('fs');
 const isDevEnv = require('./utils/misc.js').isDevEnv;
 const isMac = require('./utils/misc.js').isMac;
 const getRegistry = require('./utils/getRegistry.js');
+const ws = require('windows-shortcuts');
 
 /**
  * reads global configuration file: config/Symphony.config. this file is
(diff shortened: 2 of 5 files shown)


lneir:
please provide a description above for this commit.

config/Symphony.config
```diff
@@ -1,3 +1,5 @@
 {
     "url": "https://foundation-dev.symphony.com"
+    "launchOnStartup": true,
+    "minimizeOnClose": true
```
lneir: why is minimize on close code here, that is a separate ticket?

js/getConfig.js
```diff
@@ -50,6 +51,14 @@ var getConfig = function () {
                 }).catch(function (){
                     resolve(config);
                 });
+
+                if (config.launchOnStartup === "true"){
```
lneir: does this handle both bool and string value?

js/getConfig.js
```diff
@@ -50,6 +51,14 @@ var getConfig = function () {
...
+                    const execFile = 'Symphony.exe';
+                    let execPath = path.join(app.getAppPath(), execFile);
+                    ws.create("%APPDATA%/Microsoft/Windows/Start Menu/Programs/Startup/Symphony.lnk", execPath);
```
lneir: does this work for both install for all and install for individual user?

js/menus/menuTemplate.js
```diff
@@ -156,9 +170,68 @@ function getTemplate(app) {
...
+    {
+      // Window menu when Windows.
+        template[2].submenu.push(
+            {
+                label: 'Auto Launch on Windows Startup', 
+                type: 'checkbox', 
+                checked: launchOnStartup,
+                click: function (item) {
+                    if (item.checked){
+                        const execFile = 'Symphony.exe';
```
lneir: this looks like it is overwrite global value.  if user changes, it should only effect their user setting.  that is, if user has not set a value, then user global value, if user has modified then use the user/individual setting.

config/Symphony.config
rbarros-daitan: As the code to load/save minimizeOnClose settings is almost the same of auto launch, I have already prepared it together so ELECTRON-40 has only to implement the menu behavior itself.

PR #62, opened 2017-04-20
saved layout

lneir:
initial code that implemented "saved layout" feature.  will restore main window and pop-out windows to last positions when restarting (or refreshing the app).

note: unit tests will come in follow on PR.


lneir:
@rbarros-daitan please review code here.  the updated config read/write here should help you with your work.

lneir:
ah yes, good. thanks i'll fix that.

PR #68, opened 2017-04-23
only allow single instance of app

lneir:
only allow single instance of app to run at time.
note: this will be needed later when implementing protocol handler.


lneir:
@vikasshashidar @rbarros-daitan please review.

VikasShashidhar:
@lneir Looks good to me.

PR #69, opened 2017-04-26
electron-17: implemented crash reporter for all the renderer process …

VishwasShashidhar:
electron-17: implemented crash reporter for all the renderer process and the main process. Also, added a button on the demo page to trigger crash reporting.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -1,157 +1,176 @@
 <html>
-    <head>
-    </head>
-    <body>
-        <h1>Symphony Electron API Demo</h1>
-        <hr>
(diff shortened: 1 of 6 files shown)


VishwasShashidhar:
@lneir Could you please review this and let me know if we need to change anything?

VishwasShashidhar:
@lneir Also, electron already has unit tests covered for crash reporting as mentioned here -> https://github.com/electron/electron/blob/master/spec/api-crash-reporter-spec.js

Should we still write any custom unit tests for crash reporting?

PR #70, opened 2017-04-27
get media source interface

lneir:
This code implements interface that allows getting list of media sources in sandbox renderer process. This code provides equivalent of desktopCapturer.getSources that works in a sandbox renderer. see: https://electron.atom.io/docs/api/desktop-capturer/.


lneir:
@KiranNiranjan please review.  you'll need this PR to complete screen share.

PR #81, opened 2017-05-09
fix dialog when hitting close

lneir:
allows close button in dialog to cancel dialog rather than attempting retry.


lneir:
@KiranNiranjan please review, thanks.

PR #85, opened 2017-05-10
Electron 24 - protocol handler

VishwasShashidhar:
Implemented the protocol handler functionality on electron app for the following use cases:

1. open IM
2. open chat room
3. open user profile
4. use cases handling opening the electron symphony app in case it is already not open and loading the appropriate stream once the bridge is loaded.


VishwasShashidhar:
@lneir @rbarros-daitan  Can you review this please?

VishwasShashidhar:
I've pushed a new commit that adds support for windows. Apparently, electron doesn't emit 'open-url' event on windows. So, we've had to work around this by capturing the process argument listening on the callback of the make single instance function for when the app is already open. In case the app is not open, we listen on the 'on-ready' event and process the protocol action.

There is an open pull request on electron repo that emits the 'open-url' event even for windows, maybe once that is merged, we can use a single event to process the protocol action for both mac and windows.

More details here -> https://github.com/electron/electron/issues/9172 & https://github.com/electron/electron/pull/8052

js/main.js
```diff
@@ -50,52 +58,61 @@ var symphonyAutoLauncher = new AutoLaunch({
 app.on('ready', getUrlAndOpenMainWindow);
 
 function getUrlAndOpenMainWindow() {
+
+    if (process.platform === 'win32') {
```
lneir: why only win32 here?  please add documentation.

js/main.js
```diff
@@ -21,15 +26,18 @@ require('./memoryMonitor.js');
...
     let mainWin = windowMgr.getMainWindow();
     if (mainWin) {
+        isAppAlreadyOpen = true;
         if (mainWin.isMinimized()) {
             mainWin.restore();
         }
         mainWin.focus();
     }
+    // Handle protocol action if necessary
+    processProtocolAction(argv);
```
lneir: shouldn't this be moved to line 33?  also why this line and line 62 both?

js/main.js
```diff
@@ -146,3 +193,15 @@ app.on('activate', function () {
         windowMgr.showMainWindow();
     }
 });
+
+app.setAsDefaultProtocolClient('symphony');
+
+app.on('open-url', function (event, url) {
+    if (!isAppAlreadyOpen) {
```
lneir: shouldn't this only be hit if app is already open?

js/mainApiMgr.js
```diff
@@ -89,6 +95,11 @@ electron.ipcMain.on(apiName, (event, arg) => {
         log.setLogWindow(event.sender);
     }
 
+    if (arg.cmd === apiCmds.registerProtocolHandler) {
```
lneir: nit-picky but move this up to be after line 76 (close to other).

js/protocolHandler/index.js
```diff
@@ -0,0 +1,57 @@
+/**
```
lneir: don't need lines 1 to 3.

js/protocolHandler/index.js
```diff
@@ -0,0 +1,57 @@
...
+ * Created by vishwas on 04/05/17.
+ */
+'use strict';
+
+let protocolWindow;
+var protocolUrl;
```
lneir: why not use let?

js/main.js
VishwasShashidhar: added.

js/main.js
VishwasShashidhar: not exactly. `app.singleInstance` is called when we already have the app running, that means, we'll have the main window initialized in which the control will not go into the `if` condition. So, the `processProtocolAction` should be where it is now.

js/main.js
VishwasShashidhar: that is what I had thought as well until I started testing and found out that electron emits this event before the 'ready' event when the app is not open. So, we'll need to handle both the cases here.

js/protocolHandler/index.js
VishwasShashidhar: ya, removed it. IDE auto-generated.

js/protocolHandler/index.js
VishwasShashidhar: makes sense. I've changed it now.

VishwasShashidhar:
@lneir Please review now.

js/main.js
```diff
@@ -146,3 +203,9 @@ app.on('activate', function () {
         windowMgr.showMainWindow();
     }
 });
+
+app.setAsDefaultProtocolClient('symphony');
+
+app.on('open-url', function (event, url) {
+    handleProtocolAction(url);
```
lneir: so on mac is this fired both when app is not running and running?

lneir: also noticed in documentation here: https://electron.atom.io/docs/api/app/#event-open-url-macos that something might be needed in plist???

js/main.js
```diff
@@ -146,3 +203,9 @@ app.on('activate', function () {
         windowMgr.showMainWindow();
     }
 });
+
+app.setAsDefaultProtocolClient('symphony');
+
+app.on('open-url', function (event, url) {
```
lneir: as note that this event is only supported on mac.

js/main.js
VishwasShashidhar: yes, it is triggered for both when the app is running and not running. The plist thing is taken care of automatically by the `app.setAsDefaultProtocolClient('symphony');` call.

js/main.js
VishwasShashidhar: correct, only supported on mac currently.

PR #88, opened 2017-05-11
enforce min width and height for windows

lneir:
enforces min width of 300 and min height for 600 for main and pop-out windows as per bug noted in electron-52

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -144,7 +144,7 @@
 
         var openWinButton = document.getElementById('open-win');
         openWinButton.addEventListener('click', function() {
-            win = window.open('win.html?x=100&y=100', 'test-window', 'height=100,width=100');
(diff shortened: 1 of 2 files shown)


lneir:
@VishwasShashidhar please review.

demo/index.html
```diff
@@ -144,7 +144,7 @@
 
         var openWinButton = document.getElementById('open-win');
         openWinButton.addEventListener('click', function() {
-            win = window.open('win.html?x=100&y=100', 'test-window', 'height=100,width=100');
+            win = window.open('win.html?x=100&y=100', 'test-window', 'height=800,width=400');
```
VishwasShashidhar: since there are no specs for this and since this is a demo page, I think we should be good to go.

PR #91, opened 2017-05-18
adds getVersionInfo api

lneir:
Implements api to get version info... see api details here: https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/Version+API

### DIFF
--- a/js/preload/preloadMain.js
+++ b/js/preload/preloadMain.js
@@ -58,8 +58,19 @@ function createAPI() {
     // API exposed to renderer process.
     //
     window.ssf = {
-        // api version
-        version: '1.0.0',
+        getVersionInfo: function() {
+            return new Promise(function(resolve) {
+                var appName = remote.app.getVersion();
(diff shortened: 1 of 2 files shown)


js/preload/preloadMain.js
```diff
@@ -58,8 +58,19 @@ function createAPI() {
     // API exposed to renderer process.
     //
     window.ssf = {
-        // api version
-        version: '1.0.0',
+        getVersionInfo: function() {
+            return new Promise(function(resolve) {
+                var appName = remote.app.getVersion();
+                var appVer = remote.app.getName();
```
VishwasShashidhar: There seems to be a mix up with the variable assignment here.

lneir: yes you are correct.  thanks i'll fix.

lneir:
@VishwasShashidhar pushed change for issued you mentioned.  please review.

PR #96, opened 2017-05-22
simplify protocol handler api

lneir:
this removes checkProtocolAction from preload api  as it should not be needed anymore.  requires that corresponding PR in SFE-Cilent be merged as well: https://github.com/SymphonyOSF/SFE-Client-App/pull/5622


VishwasShashidhar:
lgtm

PR #98, opened 2017-05-23
Electron 19 (Frameless Window)

KiranNiranjan:
Added support for frameless window


KiranNiranjan:
@lneir Please review. Thanks

lneir:
please rebase PR from latest, there are many other PR mixed in here.

js/windowMgr.js
```diff
@@ -73,6 +73,7 @@ function doCreateMainWindow(initialUrl, initialBounds) {
         show: true,
         minWidth: MIN_WIDTH,
         minHeight: MIN_HEIGHT,
+        titleBarStyle: 'hidden',
```
lneir: what does this do on windows?
also looks like we loose capability to move window around?

lneir: also looks like we loose capability to move window around?  maybe change in client fixes this???

KiranNiranjan: This settings will not affect wrapper running in windows.

KiranNiranjan: I have added this ```-webkit-app-region: drag``` class to the header section in client-app which takes care of the draggable window.

lneir: but won't the client code impact?

KiranNiranjan:
Will open a new pull request.

PR #101, opened 2017-05-24
fix mac pkg installer path

lneir:
makes path relative rather than absolute.


VishwasShashidhar:
yes, sorry, this should have been taken care of. Anyway, lgtm, can be merged.

PR #104, opened 2017-05-24
Electron-17 Crash Reporter

VikasShashidhar:
1. Implemented crash reporter for both main and renderer processes
2. Fetch crash log details from Symphony.config
3. Send reports to a break pad server


VishwasShashidhar:
@VikasShashidhar Looks like your old repository doesn't exist anymore. You'll need to close this. I'll raise a separate PR for electron-17 from my repo.

PR #105, opened 2017-05-24
fix opening A tag

lneir:
opening A tag link from browser (in different hostname) should be blocked as this causes electron page to navigate to new location.  But there is a bug in electron https://github.com/electron/electron/issues/8841 that is preventing main process from using 'will-navigate' event.  So in the mean-time using 'beforeunload' event in browser process to block.

Note:  A tag with href='_blank' is allowed and already handled by 'window-open' event; will cause link to be opened in default browser.


lneir:
@VishwasShashidhar or @KiranNiranjan please review.

PR #109, opened 2017-05-25
Electron-26 (Auto Launch for Mac)

KiranNiranjan:
Updated AutoLauncher path to make it work properly in mac osx

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -7,7 +7,7 @@ const squirrelStartup = require('electron-squirrel-startup');
 const AutoLaunch = require('auto-launch');
 const urlParser = require('url');
 const { getConfigField } = require('./config.js');
-const { isDevEnv} = require('./utils/misc.js');
+const { isMac, isDevEnv } = require('./utils/misc.js');
 const protocolHandler = require('./protocolHandler');
 
 // used to check if a url was opened when the app was already open
@@ -46,7 +46,7 @@ if (shouldQuit) {
 
 var symphonyAutoLauncher = new AutoLaunch({
     name: 'Symphony',
-    path: process.execPath,
+    path: isMac ? '/Applications/' + app.getName() + '.app' : process.execPath,
 });
 
 /**
--- a/js/main.js
+++ b/js/main.js
@@ -46,7 +46,7 @@ if (shouldQuit) {
 
 var symphonyAutoLauncher = new AutoLaunch({
     name: 'Symphony',
-    path: isMac ? '/Applications/' + app.getName() + '.app' : process.execPath,
+    path: process.execPath,
 });
 
 /**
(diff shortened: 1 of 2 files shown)


KiranNiranjan:
@lneir & @VishwasShashidhar Please review. Thanks

js/main.js
```diff
@@ -46,7 +46,7 @@ if (shouldQuit) {
 
 var symphonyAutoLauncher = new AutoLaunch({
     name: 'Symphony',
-    path: process.execPath,
+    path: isMac ? '/Applications/' + app.getName() + '.app' : process.execPath,
```
lneir: we are relying on app being installed is this exact location - which it will but in case it is put elsewhere - couldn't you instead use app.getPath('exe') and figure out from there?

KiranNiranjan: By doing that i was facing this [issue](https://github.com/Teamwork/node-auto-launch/issues/28). Auto launcher verifies the path and if it fails to locate, it will defaults to the ```exe`` path.

lneir: what about this..
var appPath = app.getPath('exe').split('.app/Content')[0] + '.app';
appLauncher = new AutoLaunch({
  name: app.getName(),
  path: appPath,
});

KiranNiranjan:
@lneir Changed the implementation. Please review. Thanks.

js/menus/menuTemplate.js
```diff
@@ -184,17 +186,35 @@ function getTemplate(app) {
             checked: launchOnStartup,
             click: function (item) {
                 if (item.checked){
-                    symphonyAutoLauncher.enable()
-                    .catch(function (err) {
-                        let title = 'Error setting AutoLaunch configuration';
-                        electron.dialog.showErrorBox(title, title + ': ' + err);
-                    });
+                    if (isMac){
+                        childProcess.exec(`launchctl load ${launchAgentPath}`, (err) => {
```
lneir: why do we do different work to change how to enable/disable startup for mac?  doesn't the auto-launch package work for mac?  i thought it should take care of all of this???

KiranNiranjan: As informed earlier in this ticket there are open issue for MacOS, we've built a custom solution. Please check the ticket here -> [issue](https://github.com/Teamwork/node-auto-launch/issues/28)

lneir: ok, then please add documentation here why we are using different soln (i.e., bug in electron and provide ref to bug - so we can track when it gets fixed).

KiranNiranjan:
Need to change this custom implementation to use AutoLaunch once this issue is fixed [#28](https://github.com/Teamwork/node-auto-launch/issues/28)

lneir:
lgtm

PR #112, opened 2017-05-26
Add Remove Previous Version action during install

rbarros-daitan:
Applying same change from DES-14034.


rbarros-daitan:
@lneir , please review it.

lneir:
lgtm, thanks.

PR #113, opened 2017-05-26
throttle should check for valid args

lneir:
throttle should throw exception if invalid args are provided.


lneir:
@KiranNiranjan please review and see my comment in chat regarding activity detection issue.

PR #114, opened 2017-05-27
Electron-32 (Activity Detection)

KiranNiranjan:
Initiating activity detection while setting up the window.


KiranNiranjan:
@lneir Please review. Thanks

PR #116, opened 2017-05-30
Electron 44 File Download Experience

VikasShashidhar:
1. Initiate download manager when a file download is initiated.
2. Add items to download manager when new items are downloaded.
3. Allow user to open file in default OS app.
4. Allow user to show file in finder/explorer.
5. Allow user to close download bar.

@lneir Please review.


js/preload/preloadMain.js
```diff
@@ -12,6 +12,7 @@
 // https://github.com/electron/electron/issues/2984
 //
 const { ipcRenderer, remote } = require('electron');
+const _ = require('underscore');
```
lneir: i really want to avoid bringing in underscore dependency?  just for findindex.  we are using very modern JS here and you can use Array.findIndex here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex?v=example

js/preload/preloadMain.js
```diff
@@ -121,6 +123,28 @@ function createAPI() {
         },
 
         /**
+         * Open file in default app.
+         */
+        openFile: function (id) {
+            let fileIndex = _.findIndex(local.downloadItems, {_id: id});
```
lneir: what happens if index is not found?  index will be -1

js/preload/preloadMain.js
```diff
@@ -121,6 +123,28 @@ function createAPI() {
...
+                remote.dialog.showErrorBox("File not found", 'The file you are trying to open cannot be found in the specified path');
+            }
+        },
+
+        /**
+         * Show downloaded file in explorer or finder.
+         */
+        showInFinder: function (id) {
+            let showFileIndex = _.findIndex(local.downloadItems, {_id: id});
+            let showResponse = remote.shell.showItemInFolder(local.downloadItems[showFileIndex].filePath);
```
lneir: again, what if index is -1 (i.e., not found)

js/preload/preloadMain.js
```diff
@@ -246,6 +270,121 @@ function createAPI() {
         }
     });
 
+    // listen for file download complete event
+    local.ipcRenderer.on('download-completed', (event, arg) => {
+
+        local.downloadItems.push(arg);
+        let downloadItemKey = arg._id;
```
lneir: add safety check in case arg._id is not found or invalid.

js/preload/preloadMain.js
```diff
@@ -246,6 +270,121 @@ function createAPI() {
...
+
+        let ul = document.getElementById('download-main');
```
lneir: add safety check, in case ul not found.

js/preload/preloadMain.js
```diff
@@ -121,6 +123,28 @@ function createAPI() {
         },
 
         /**
+         * Open file in default app.
+         */
+        openFile: function (id) {
```
lneir: why do we need this in the ssf namespace.  this will expose api to JS.  i don't think needed here?

js/preload/preloadMain.js
```diff
@@ -121,6 +123,28 @@ function createAPI() {
...
+            let openResponse = remote.shell.openExternal(`file:///${local.downloadItems[fileIndex].filePath}`);
+            if (!openResponse) {
+                remote.dialog.showErrorBox("File not found", 'The file you are trying to open cannot be found in the specified path');
+            }
+        },
+
+        /**
+         * Show downloaded file in explorer or finder.
+         */
+        showInFinder: function (id) {
```
lneir: why do we need this in the ssf namespace.  this will expose api to JS.  i don't think needed here?

js/windowMgr.js
```diff
@@ -174,6 +175,28 @@ function doCreateMainWindow(initialUrl, initialBounds) {
...
+                let data = {
+                    _id: getGuid(),
+                    filePath: item.getSavePath(),
+                    total: filesize(item.getTotalBytes()),
+                    fileName: item.getFilename(),
+                    mimeType: item.getMimeType()
+                };
+                mainWindow.send('download-completed', data);
+            } else {
+                console.log(`Download failed: ${state}`);
```
lneir: if download fails, don't we have to send something to renderer to show some failed state?  will it keep downloading forever if we don't?

package.json
```diff
@@ -92,7 +92,10 @@
     "electron-rebuild": "^1.5.7",
     "electron-squirrel-startup": "^1.0.0",
     "keymirror": "0.1.1",
-    "winreg": "^1.2.3"
+    "winreg": "^1.2.3",
+    "electron-dl": "^1.9.0",
+    "filesize": "^3.5.10",
+    "underscore": "^1.8.3"
```
lneir: remove

js/windowMgr.js
```diff
@@ -174,6 +175,28 @@ function doCreateMainWindow(initialUrl, initialBounds) {
 
     mainWindow.on('closed', destroyAllWindows);
 
+    // Manage File Downloads
+    mainWindow.webContents.session.on('will-download', (event, item, webContents) => {
```
lneir: how are we handling downloads from pop-outs?

js/preload/preloadMain.js
VikasShashidhar: Sure. Makes sense. The idea behind using underscore was to keep it ready if it was required in the future. I've changed it now.

js/preload/preloadMain.js
VikasShashidhar: I had skipped this because download items are being controlled by us. Either way, it makes sense to have a safety check. I've added it now.

js/preload/preloadMain.js
VikasShashidhar: Same as above.

js/preload/preloadMain.js
VikasShashidhar: Same as above.

js/preload/preloadMain.js
VikasShashidhar: Done.

js/preload/preloadMain.js
VikasShashidhar: Done.

js/preload/preloadMain.js
VikasShashidhar: Same as above.

js/windowMgr.js
VikasShashidhar: This is a little tricky to handle because of the way downloads work on the client app. There is now way for us to know on the Electron side whether the file download has failed. We can know when it succeeds, but hard to know whether it has failed. Will discuss more on this on call. I'll remove the else though (that is probably not required).

package.json
VikasShashidhar: Done

js/windowMgr.js
VikasShashidhar: *no way

js/windowMgr.js
VikasShashidhar: Good catch. Have handled it now for all popouts (including help, IMs and Rooms). One case I am unsure about is how do we handle downloads for 3rd party app poputs (ChartIQ for ex.). Do we use our manager for that as well? Please share your thoughts on this.

screen shot 2017-05-31 at 2 47 06 pm

lneir: good question about 3rd party... currently don't believe we have any support or need there.  but i'll touch base with app group.  for now we can skip as i know we don't support in current wrapper.

js/preload/preloadMain.js
```diff
@@ -213,6 +214,36 @@ function createAPI() {
 
     };
 
+    /**
+     * Open file in default app.
+     */
+    function openFile(id) {
```
lneir: also i to make things a bit cleaner, can we move all of this code here related to download file into a separate module.  then require it from preload.js.

VikasShashidhar: 👍

js/windowMgr.js
VikasShashidhar: Ok

js/preload/preloadMain.js
```diff
@@ -19,11 +19,14 @@ const apiCmds = apiEnums.cmds;
...
 
+require('../downloadManager/downloadManager');
+
 const nodeURL = require('url');
 
 // hold ref so doesn't get GC'ed
 const local = {
-    ipcRenderer: ipcRenderer
+    ipcRenderer: ipcRenderer,
+    downloadItems: []
```
lneir: don't think we need donwloadItems here???

lneir:
one question before merging this.... the client code is only on dev.  what happens if a user is on some other branch (say corporate) where?  it should just use default behavior - can you verify that this works ok and doesn't crash?

VikasShashidhar:
@lneir Yes. We create the necessary download manager DOM only when we have our download manager footer available. If not, it just ignores it.

https://github.com/VikasShashidhar/SymphonyElectron/blob/electron-44/js/downloadManager/downloadManager.js#L58 and https://github.com/VikasShashidhar/SymphonyElectron/blob/electron-44/js/downloadManager/downloadManager.js#L156 verifies that.

PR #117, opened 2017-05-30
add logging and unit tests

lneir:
adds updates to logging system:
- add more logging in various places in code
- adds queueing mechanism in case log messages are received before log is registered by js.
- adds unit tests for log
- adds command line option to allow changing log level: --logLevel='DEBUG' (for example).
- adds command line option to show all log messages in console: --enableConsoleLogging

also review corresponding PR in client code: https://github.com/SymphonyOSF/SFE-Client-App/pull/5696


VishwasShashidhar:
@lneir I think everything looks good. A couple of suggestions:

1. We should have a separate file for storing the messages. Maybe a json file? It'll help with localisation in the future.

2. We should try to use AOP for logging as a means for separation of concerns. This keeps the code clean. A couple of referrals to AOP in JS -> http://blog.mgechev.com/2015/07/29/aspect-oriented-programming-javascript-aop-js/, http://know.cujojs.com/tutorials/aop/intro-to-aspect-oriented-programming

Maybe both of the above are an overkill at this point in time, but definitely worth a thought.

lneir:
yes agree about localization and separate file.  but logging messages wouldn't be localized.  certainly others would that are directly displayed to user.  i'll create a separate ticket to look into localization.

lneir:
this AOP is good.  let's think about it.  yes probably for now might be overkill but we should discuss.

lneir:
@VishwasShashidhar correctly the spelling you recommended (fyi, we have been using notf as short for notification) but spelled out to explicit.  thanks.

VishwasShashidhar:
lgtm

PR #119, opened 2017-05-31
ignore single app instance for dev mode

lneir:
for development environment mode - ignore singleton app mode.  for example, this allows dev env testing and running production at the same time.


VikasShashidhar:
LGTM 👍

PR #120, opened 2017-06-01
Electron-66 (Unit tests)

KiranNiranjan:
Added more unit tests for config, getRegistry & throttle.

@lneir Please review. Thanks


tests/utils/getRegistry.test.js
```diff
@@ -0,0 +1,66 @@
+const getRegistry = require('../../js/utils/getRegistry.js');
+
+const { isMac } = require('../../js/utils/misc.js');
+
+describe('Tests for getRegistry', function() {
+
+    describe('Should not get registry for mac', function() {
+
+        it('should fail to get path for mac', function(done) {
```
lneir: what happens when we run this test on windows?  seems like it would fail?

tests/activityDetection.test.js
```diff
@@ -1,10 +1,30 @@
...
 
-    beforeAll(function () {
-        activityDetection.setActivityWindow(120000, electron.ipcRenderer);
+describe('Tests for Activity Detection', function() {
+
+    var originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 50000;
+
+    beforeAll(function (done) {
+        childProcess.exec('npm rebuild --runtime=electron --target=1.2.2 --disturl=https://atom.io/download/atom-shell --build-from-source', function (err) {
```
lneir: that target works?  even though our electron version is 1.6.11???  have you run all unit tests locally on mac and windows?

KiranNiranjan: Yep it works. Tested both on mac and windows.

PR #121, opened 2017-06-01
Electron-65 (Logging)

KiranNiranjan:
Added more logs

@lneir Please review. Thanks


js/config.js
```diff
@@ -41,6 +41,7 @@ function getUserConfigField(fieldName) {
             return config[fieldName];
         }
 
+        log.send(logLevels.ERROR, 'config: field does not exist in user config: ' + fieldName);
```
lneir: i don't think we want an error log here, in certain cases it legitimately will only be in global config, so this might be hit.  error handling for this cases should be at a higher level where it doesn't appear in either.

js/config.js
```diff
@@ -51,13 +52,15 @@ function readUserConfig() {
 
         fs.readFile(configPath, 'utf8', function(err, data) {
             if (err) {
+                log.send(logLevels.ERROR, 'config: cannot open user config file: ' + configPath + ', error: ' + err);
```
lneir: i don't think we need all of these log message where we already doing a promise reject.  this should be handled at a higher level.

ditto for all the other below here in this file.

what do others think?

js/desktopCapturer/getSources.js
```diff
@@ -29,6 +32,9 @@ function isValid(options) {
 function getSources(options, callback) {
     var captureScreen, captureWindow, id;
     if (!isValid(options)) {
+        log.send(logLevels.ERROR, 'getSources: Error getSource() ' +
```
lneir: same comment here as above.

js/screenSnippet/ScreenSnippet.js
```diff
@@ -93,9 +94,11 @@ class ScreenSnippet {
                     // no such file exists, user likely aborted
                     // creating snippet. also include any error when
                     // creating child process.
+                    log.send(logLevels.ERROR, 'ScreenSnippet: ENOENT file does not exist: '+ childProcessErr);
```
lneir: don't think needed as promise as is already return error

js/screenSnippet/ScreenSnippet.js
```diff
@@ -93,9 +94,11 @@ class ScreenSnippet {
...
                     returnErr = this._createWarn('file does not exist ' +
                         childProcessErr);
                 } else {
+                    log.send(logLevels.ERROR, 'ScreenSnippet: ERROR: ' + readErr + ',' + childProcessErr);
```
lneir: ditto above comment

js/screenSnippet/ScreenSnippet.js
```diff
@@ -74,6 +74,7 @@ class ScreenSnippet {
             child = childProcess.execFile(captureUtil, captureUtilArgs, (error) => {
                 // will be called when child process exits.
                 if (error && error.killed) {
+                    log.send(logLevels.ERROR, 'ScreenSnippet: ERROR executing command: '+ error);
```
lneir: don't think needed this is legitimate, as we kill any previous process. so only one process at a time.

js/screenSnippet/ScreenSnippet.js
```diff
@@ -105,6 +108,7 @@ class ScreenSnippet {
             }
 
             if (!data) {
+                log.send(logLevels.ERROR, 'ScreenSnippet: no file data provided');
```
lneir: shouldn't need as reject already has error.

js/screenSnippet/ScreenSnippet.js
```diff
@@ -117,20 +121,18 @@ class ScreenSnippet {
                     data: output
                 });
             } catch (error) {
+                log.send(logLevels.ERROR, 'ScreenSnippet: Buffer Error: '+ error);
```
lneir: ditto

js/utils/getRegistry.js
```diff
@@ -25,20 +27,23 @@ var symphonyRegistryHKLM6432 = new Registry({
 var getRegistry = function (name) {
     var promise = new Promise(function(resolve, reject) {
         if (isMac) {
+            log.send(logLevels.ERROR, 'getRegistry: Registry is not supported for mac osx.');
```
lneir: not needed.

js/utils/throttle.js
```diff
@@ -7,9 +9,11 @@
  */
 function throttle(throttleTime, func) {
     if (typeof throttleTime !== 'number' || throttleTime <= 0) {
+        log.send(logLevels.ERROR, 'throttle: invalid throttleTime arg, must be a number: ' + throttleTime);
```
lneir: already has throw.

js/utils/throttle.js
```diff
@@ -7,9 +9,11 @@
...
         throw Error('throttle: invalid throttleTime arg, must be a number: ' + throttleTime);
     }
     if (typeof func !== 'function') {
+        log.send(logLevels.ERROR, 'throttle: invalid func arg, must be a function: ' + func);
```
lneir: already has throw.

PR #124, opened 2017-06-02
ELECTRON-66 (Unit tests)

KiranNiranjan:
Fixed the target version


KiranNiranjan:
@lneir Please review.

lneir:
this will work for now as we have 7.10.0 on our build servers.  but is there a way to make this dynamically configurable?

PR #126, opened 2017-06-05
electron-69: sets minimize on close to false

VishwasShashidhar:
fixes the issue raised in electron-69


lneir:
lgtm.

you'll also need another pr to fix this for windows installer.

VishwasShashidhar:
will work on it.

PR #127, opened 2017-06-05
Electron-56 (Always on top)

KiranNiranjan:
Implemented always on top feature.

@lneir Please review. Thanks


js/windowMgr.js
```diff
@@ -31,6 +31,9 @@ let windows = {};
 let willQuitApp = false;
 let isOnline = true;
 let boundsChangeWindow;
+let alwaysOnTop = false;
+// Get user preference for always on top
+getUserPreference();
```
lneir: the method is async.  so i don't see how this will work reliably

js/windowMgr.js
```diff
@@ -186,6 +195,9 @@ function doCreateMainWindow(initialUrl, initialBounds) {
     // open external links in default browser - a tag with href='_blank' or window.open
     mainWindow.webContents.on('new-window', function (event, newWinUrl,
                                                       frameName, disposition, newWinOptions) {
+
+        // Get user preference for always on top
+        getUserPreference();
```
lneir: ditto async comment above.

js/menus/menuTemplate.js
```diff
@@ -232,12 +233,29 @@ function getTemplate(app) {
...
-    // Window menu -> minimizeOnClose.
+    // Window menu -> alwaysOnTop.
+    template[index].submenu.push(
+        {
+            label: 'Always on top',
+            type: 'checkbox',
+            checked: alwaysOnTop,
+            click: (item) => {
+                alwaysOnTop = item.checked;
+                let browserWin = electron.BrowserWindow.getAllWindows();
```
lneir: i see this as functionality that should be provided by windowMgr code.  we should see about moving that there, but i see this would cause circular dep, so need to think about how to do this?  maybe some refactoring needed.

lneir:
status here?  @KiranNiranjan

js/menus/menuTemplate.js
```diff
@@ -268,14 +284,22 @@ function setCheckboxValues(){
...
     getConfigField('launchOnStartup').then(function(lStartup) {
         launchOnStartup = lStartup;
     }).catch(function (err){
         let title = 'Error loading configuration';
         log.send(logLevels.ERROR, 'MenuTemplate: error getting config field launchOnStartup, error: ' + err);
         electron.dialog.showErrorBox(title, title + ': ' + err);
     });
+
+    getConfigField('alwaysOnTop').then(function(mAlwaysOnTop) {
+        alwaysOnTop = mAlwaysOnTop;
```
lneir: why not publish event change here.  then use in windowMgr.js as value.

js/windowMgr.js
```diff
@@ -386,6 +419,24 @@ function openUrlInDefaultBrower(urlToOpen) {
     }
 }
 
+/**
+ * Called when an event is received from menu
+ * @param boolean weather to enable or disable alwaysOnTop.
+ */
+function setAlwaysOnTop(boolean) {
+    let browserWins = BrowserWindow.getAllWindows();
```
lneir: also set local value 'alwaysOnTop' value here.  then use above.

js/menus/menuTemplate.js
KiranNiranjan: Done

js/windowMgr.js
```diff
@@ -386,6 +397,25 @@ function openUrlInDefaultBrower(urlToOpen) {
...
+    if (browserWins.length > 0) {
+        browserWins.forEach(function (browser) {
+            browser.setAlwaysOnTop(boolean);
+        });
+    }
+}
+
+// node event emitter to update always on top
+eventEmitter.on('alwaysOnTop', (boolean) => {
```
lneir: let's be a little more descriptive with name: isAlwaysOnTop

js/windowMgr.js
```diff
@@ -386,6 +397,25 @@ function openUrlInDefaultBrower(urlToOpen) {
     }
 }
 
+/**
+ * Called when an event is received from menu
+ * @param boolean weather to enable or disable alwaysOnTop.
+ */
+function setAlwaysOnTop(boolean) {
```
lneir: add more descriptive name, maybe: isAlwaysOnTop

KiranNiranjan:
Done. @lneir Please review.

js/eventEmitter.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
```
lneir: rather than exposing entire EventEmitter interface, just expose necessary methods...

module.exports = {
   emit: eventEmitter.emit,
   on: eventEmitter.on
}

also make clear in documentation that this can only be used main process.

i would like a better name but haven't thought of one yet.

KiranNiranjan: @lneir changed the exports. I didn't quite understand ```i would like a better name but haven't thought of one yet.```

PR #133, opened 2017-06-07
fix activity detection tests and build process

lneir:
when building activity detection for tests need to provide node version (e.g., v7.4.0) since we are building for jest env which is running in local version of node.

also re-organized build steps to ensure electron-rebuild is always run first.  this is needed since running tests will build a version that is incompatible with electron.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,21 +6,17 @@
   "author": "Symphony",
   "main": "js/main.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


lneir:
@KiranNiranjan please review this.  this gets it working correctly, should work now on build servers and local.

KiranNiranjan:
lgtm 👍

PR #134, opened 2017-06-08
electron-69: fixes the issue raised in the ticket for windows

VishwasShashidhar:

### DIFF
--- a/installer/win/Symphony-x64.aip
+++ b/installer/win/Symphony-x64.aip
@@ -20,7 +20,6 @@
     <ROW Property="BannerBitmap" Value="banner" MultiBuildValue="DefaultBuild:Banner.jpg" Type="1" MsiKey="BannerBitmap"/>
     <ROW Property="CTRLS" Value="2"/>
     <ROW Property="DialogBitmap" Value="dialog" MultiBuildValue="DefaultBuild:Tabloid.jpg" Type="1" MsiKey="DialogBitmap"/>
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
     "electron-builder": "^13.9.0",
     "electron-builder-squirrel-windows": "^12.3.0",
     "electron-packager": "^8.5.2",
+    "electron-rebuild": "^1.5.7",
```
VishwasShashidhar: I've moved this dependency to dev-dependencies as recommended by electron-builder

installer/win/Symphony-x64.aip
```diff
@@ -20,7 +20,6 @@
     <ROW Property="BannerBitmap" Value="banner" MultiBuildValue="DefaultBuild:Banner.jpg" Type="1" MsiKey="BannerBitmap"/>
     <ROW Property="CTRLS" Value="2"/>
     <ROW Property="DialogBitmap" Value="dialog" MultiBuildValue="DefaultBuild:Tabloid.jpg" Type="1" MsiKey="DialogBitmap"/>
-    <ROW Property="MINIMIZE_ON_CLOSE" Value="CheckBox"/>
```
lneir: why are we getting rid of property, i thought we just wanted to change default from true to false, on line 153.

VishwasShashidhar: Apparently, with advanced installer, if the checkbox is selected, it adds this row and sets the value as "Checkbox", nothing of the sort of true / false.

PR #139, opened 2017-06-12
cache config files when read from disk

lneir:
adds in memory caching of user and global config file when read from disk.  will keep from hitting disk on every read.

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -61,17 +61,14 @@ function readUserConfig() {
             if (err) {
                 reject('cannot open user config file: ' + configPath + ', error: ' + err);
             } else {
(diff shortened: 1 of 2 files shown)


js/config.js
```diff
@@ -158,15 +181,24 @@ function saveUserConfig(fieldName, newValue, oldConfig) {
             if (err) {
                 reject(err);
             } else {
+                userConfig = newConfig;
                 resolve(newConfig);
```
lneir: @VishwasShashidhar pushed update, please review.

VishwasShashidhar:
lgtm

PR #140, opened 2017-06-13
electron-75 -> Fixes the issue

VishwasShashidhar:


lneir:
what is issue is this addressing?  looks like some dir doesn't exist???

VishwasShashidhar:
yes, so, by default, the `LaunchAgents` directory doesn't exist when we install the OS afresh. Hence creating this in case it doesn't exist.

PR #142, opened 2017-06-14
Electron-43 (Configure Notification Position)

KiranNiranjan:
Implemented configure notification position.

@lneir & @VikasShashidhar Please review. Thanks


config/Symphony.config
```diff
@@ -1,5 +1,9 @@
 {
     "url": "https://foundation-dev.symphony.com",
     "minimizeOnClose" : false,
-    "launchOnStartup" : true
+    "launchOnStartup" : true,
+    "notfSettings" : {
```
VikasShashidhar: Can you rename this to "notifications"?

config/Symphony.config
```diff
@@ -1,5 +1,9 @@
...
+        "notfPosition" : "upper-right",
```
VikasShashidhar: Just "position" is cleaner.

config/Symphony.config
```diff
@@ -1,5 +1,9 @@
...
+        "notfDisplay" : "1234"
```
VikasShashidhar: Only "display" please.

js/enums/api.js
```diff
@@ -11,6 +11,7 @@ const cmds = keyMirror({
     registerBoundsChange: null,
     registerProtocolHandler: null,
     registerActivityDetection: null,
+    showAlertSettings: null,
```
VikasShashidhar: Can we change the use of alert to notifications?

@lneir I have observed that in some places we use "Alerts" and in some places we use "Notifications". I think we should stick with either one of them for consistency sakes.

js/mainApiMgr.js
```diff
@@ -12,6 +12,7 @@ const logLevels = require('./enums/logLevels');
 const activityDetection = require('./activityDetection/activityDetection');
 const badgeCount = require('./badgeCount.js');
 const protocolHandler = require('./protocolHandler');
+const configureAlert = require('./notify/settings/configure-alert-position');
```
VikasShashidhar: Same as above. Use either "Alerts" or "Notifications".

js/mainApiMgr.js
```diff
@@ -98,6 +99,10 @@ electron.ipcMain.on(apiName, (event, arg) => {
         // renderer window that has a registered activity detection from JS.
         activityDetection.setActivityWindow(arg.period, event.sender);
     }
+
+    if (arg.cmd === apiCmds.showAlertSettings) {
+        configureAlert.openConfigurationWindow();
```
VikasShashidhar: Can you rename this to "openNotificationConfiguration" or "openAlertConfiguration"?

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
+html, body {
+    background: #F9F9FA;
+    margin: 0;
+    height: 100%;
+    font-family: SymphonyLato, Lato, serif;
```
VikasShashidhar: Please default this to sans-serif. Let the OS decide what sans-serif font to use. Let the opened configuration window have as close to native styles as possible.

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
...
+
+.radio__label {
+    cursor: pointer;
+}
+
+.footer {
+    padding: 8px 16px;
+    border-top: 1px solid rgba(0, 0, 0, .1);
+    display: -webkit-flex;
+    display: -ms-flexbox;
```
VikasShashidhar: Not required. App always runs inside a chromium wrapper.

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
...
+    display: flex;
+    -webkit-align-items: center;
+    -ms-flex-align: center;
```
VikasShashidhar: Not required. App always runs inside a chromium wrapper.

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
...
+    margin-left: auto;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-align-items: center;
+    -ms-flex-align: center;
+    align-items: center;
+}
+
+.button {
```
VikasShashidhar: Let's remove this entirely. Let the OS give its styles.

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
...
+    line-height: 1;
+    user-select: none;
+    overflow: hidden;
+    position: relative;
+    transition: all 250ms ease-in-out;
+    z-index: 1;
+    text-transform: uppercase;
+}
+
+.buttonDismiss {
```
VikasShashidhar: Same with this. Please remove.

js/notify/settings/configure-alert-position.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
```
VikasShashidhar: I know we won't have the title showing up in our child window, but can you please change this to "Configure Notification Position"?

js/notify/settings/configure-alert-position.js
```diff
@@ -0,0 +1,148 @@
...
+let configurationWindow;
+let screens;
+let notfPosition;
+let notfDisplay;
+
+let windowConfig = {
+    width: 460,
+    height: 360,
+    alwaysOnTop: true,
+    skipTaskbar: true,
```
VikasShashidhar: Would we not want this to show up in the taskbar just in case the user switches back to the main window for some reason?

js/enums/api.js
lneir: internally let's stick with name notifications.

js/notify/settings/configure-alert-position.css
```diff
@@ -0,0 +1,143 @@
...
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-align-items: center;
+    -ms-flex-align: center;
+    align-items: center;
+}
+
+.buttonLayout {
+    margin-left: auto;
+    display: -webkit-flex;
```
lneir: no need for prefix on flex, can use only display: flex.

ditto comment on other prefixes here.  we are in a known env.

js/notify/settings/configure-alert-position.js
```diff
@@ -0,0 +1,148 @@
...
+
+const { getConfigField, updateConfigField } = require('../../config');
+
+let configurationWindow;
+let screens;
+let notfPosition;
+let notfDisplay;
+
+let windowConfig = {
+    width: 460,
```
lneir: make this a modal: https://electron.atom.io/docs/api/browser-window/#modal-windows

js/notify/settings/configure-alert-position.js
```diff
@@ -0,0 +1,148 @@
...
+    try {
+        fs.statSync(templatePath).isFile();
+    } catch (err) {
+        log.send(logLevels.ERROR, 'configure-alert-position: Could not find template ("' + templatePath + '").');
+    }
+    return 'file://' + templatePath;
+}
+
+function openConfigurationWindow() {
+    let mainWin = configurationWindow;
```
lneir: probably shouldn't need some of this code here if we make it a modal.

js/notify/settings/configure-alert-position.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+function createWindow() {
+    configurationWindow = new BrowserWindow(windowConfig);
+    configurationWindow.setVisibleOnAllWorkspaces(true);
+    configurationWindow.loadURL(getTemplatePath());
+    configurationWindow.show();
+
+    configurationWindow.webContents.on('did-finish-load', function () {
+        configurationWindow.webContents.send('screens', screens);
```
lneir: screens could change at any time.  looks like using value at app load.

js/notify/electron-notify.js
```diff
@@ -208,12 +227,32 @@ function calcDimensions() {
...
 
-    // Use primary display only
-    let display = electron.screen.getPrimaryDisplay();
+    // This feature only applies to windows
+    if (!isMac) {
+        let screens = electron.screen.getAllDisplays();
+        if (screens && screens.length >= 0) {
+            externalDisplay = screens.find((screen) => {
+                let screenId = screen.id.toString();
+                return ((screen.bounds.x !== 0 || screen.bounds.y !== 0) && screenId === notfDisplay);
```
lneir: so you are trying to check that an external display with given id exists?

lneir: why check the bounds?

js/notify/electron-notify.js
```diff
@@ -208,12 +227,32 @@ function calcDimensions() {
...
+        // use user selected display
+        display = externalDisplay;
+        config.corner = {};
+        config.corner.x = display.bounds.x;
+        config.corner.y = display.bounds.y;
+    } else {
+        // Use primary display
+        display = electron.screen.getPrimaryDisplay();
+        config.corner = {};
+        config.corner.x = display.bounds.x + display.workArea.x;
```
lneir: don't we also want the same value here for external display?
that is code could be simplified:
```
config.corner = {}
config.corner.x = display.bounds.x + display.workArea.x;
config.corner.y = display.bounds.y + display.workArea.y;
let display = externalDisplay ? externalDisplay : electron.screen.getPrimaryDisplay();
```

js/notify/settings/configure-alert-position.js
```diff
@@ -0,0 +1,148 @@
...
+    };
+    updateConfigField('notfSettings', notfSettings);
+    updateNotification(notfPosition, notfDisplay);
+}
+
+/**
+ * Method to read 'notfPosition' & 'notfDisplay' from config and
+ * updated the configuration view
+ */
+function loadConfig() {
```
lneir: why do we have to loadConfig here?  can't we get via event?

js/notify/electron-notify.js
KiranNiranjan: For external display we should not be adding the ```display.workArea``` because the ```display.workArea.x``` value includes the width of the ```PrimaryDisplay```.

> **Primary Display**
bounds: { x: 0, y: 0, width: 1280, height: 800 }
workArea: { x: 0, y: 23, width: 1280, height: 737 }
workAreaSize: { width: 1280, height: 737 }

> **External Display**
bounds: { x: 1280, y: 0, width: 1920, height: 1080 }
workArea: { x: 1280, y: 23, width: 1920, height: 1057 }
workAreaSize: { width: 1920, height: 1057 }

As you can see if we add ```display.workArea``` the notification window's ```corner.x``` value would be ```bounds : 1280 + workArea : 1280```.

So i think this code cannot be simplified because the external display should be handled separately.

js/notify/electron-notify.js
KiranNiranjan: Because we need only the external display & not primary display.

Please look at my below comment for more info.

KiranNiranjan:
@lneir & @VikasShashidhar Refactored the code. Please review. Thanks

config/Symphony.config
```diff
@@ -1,5 +1,9 @@
 {
     "url": "https://foundation-dev.symphony.com",
     "minimizeOnClose" : false,
-    "launchOnStartup" : true
+    "launchOnStartup" : true,
+    "notifications": {
+        "position": "upper-right",
+        "display": "1234"
```
VikasShashidhar: Sorry, why do we have a 1234 here? Is this a dummy value? If so, why not keep it empty instead?

js/preload/preloadMain.js
```diff
@@ -209,7 +209,15 @@ function createAPI() {
          * see: https://electron.atom.io/docs/api/desktop-capturer/
          * for interface: see documentation in desktopCapturer/getSources.js
          */
-        getMediaSources: getMediaSources
+        getMediaSources: getMediaSources,
+
+        showAlertSettings: function() {
```
VikasShashidhar: Sorry, I am nitpicking a bit here. We still want this to be "showNotificationSettings". Consistency please :)

KiranNiranjan: Changing this method here i'll also need to update the client app. Shall i make this change and create a PR in client app?

config/Symphony.config
lneir: i would assume this should be empty and would use primary display.

js/notify/electron-notify.js
lneir: ok, then can't we just use display.workArea for both?

lneir:
checked out branch to play with and noticed some issues:
- config name of notification is different than what coding is expecting
- for multi-monitor - doesn't seem to work properly - always appears on primary monitor
- when hooked to external monitor with laptop closed, then when opening laptop to get two monitors i get this error: 
screen shot 2017-06-15 at 10 48 34 am

config/Symphony.config
KiranNiranjan: Done

js/preload/preloadMain.js
KiranNiranjan: Done

js/notify/electron-notify.js
```diff
@@ -208,12 +227,21 @@ function calcDimensions() {
 function setupConfig() {
     closeAll();
 
-    // Use primary display only
-    let display = electron.screen.getPrimaryDisplay();
+    // This feature only applies to windows
+    if (!isMac) {
```
KiranNiranjan: @lneir Fixed ```first``` & ```third``` points. For the ```second``` point this feature is restricted to work only on windows.

PR #146, opened 2017-06-21
Adding always on top tests.

VikasShashidhar:
Please note: Some of these tests only run when the sandbox mode is turned off as we get access to browserWindow only then.

@lneir Please review and share initial feedback. Some of this needs to be fine-tuned and I am working on it. Meanwhile, thought you'd want to see this.


VikasShashidhar:
Please note: Not ready for a merge yet.

VikasShashidhar:
@lneir I've ported these tests to use Jest and have removed chai and mocha. It works fine for two tests but there are problems for tests involving the browserWindow object as it's not available when nodeIntegration is turned off. Unfortunately, there isn't a simple way to resolve this and I am looking at how to resolve this.

I have some leads -> https://github.com/electron/spectron/issues/54, https://github.com/electron/spectron#node-integration

I tried the recommended solution in the second link, but for some reason, I am not able to get it to work.

I'll continue looking for solutions. Meanwhile, I've made some commits to this.

PR #147, opened 2017-06-25
Electron 48 : launch app fully sandboxed

VishwasShashidhar:

CC: @VikasShashidhar


VishwasShashidhar:
FYI, Symphony.exe is the executable (created from the batch file) that opens our actual electron executable.

VishwasShashidhar:
Changes w.r.t. auto-updates will be pushed in the electron-16 branch

VishwasShashidhar:
This PR includes changes for both mac and windows.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Symphony",
-  "productName": "Symphony",
+  "productName": "Symphony-Electron",
```
VikasShashidhar: I think this depends on Symphony's branding strategy for the Electron App and on whether the existing Desktop Wrapper will be continued to be used after the Electron wrapper has been released. We can probably keep this as "Symphony" only.

Same in other places as well.

@lneir Please share your inputs.

VishwasShashidhar:
@lneir @VikasShashidhar 

I've fixed the product name issue raised above now. Instead, we would now rename the generated exe as Symphony-Electron.exe.

Note that this will not be the user facing exe, it'll be the binary launched from the batch file (converted to exe).

Please review now.

VikasShashidhar:
LGTM 👍

PR #154, opened 2017-07-06
Refactored spectron tests

KiranNiranjan:
@VikasShashidhar Please review. Thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,9 @@
     "prebuild": "npm run rebuild && npm run browserify-preload",
     "browserify-preload": "browserify -o js/preload/_preloadMain.js -x electron --insert-global-vars=__filename,__dirname js/preload/preloadMain.js",
     "rebuild": "electron-rebuild -f",
-    "test": "npm run lint && jest --verbose --testPathPattern test",
+    "test": "npm run lint && npm run copy-config && jest --verbose --testPathPattern test --runInBand",
     "lint": "eslint --ext .js js/",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -79,6 +80,7 @@
     "eslint-plugin-jsx-a11y": "^4.0.0",
     "eslint-plugin-react": "^6.10.0",
     "jest": "^19.0.2",
+    "ncp": "^2.0.0",
```
VikasShashidhar: Is there no other way we can do this? I mean, copying the config file to the required path. If there is no other way, then I'll merge this.

tests/bringToFront.test.js
```diff
@@ -1,56 +1,83 @@
-const Application = require('./spectronSetup');
-
 describe('Tests for Bring to front', () => {
 
     let originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
-    jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
```
VikasShashidhar: Can we get these from a method in spectronSetup instead of hardcoding it here?

package.json
KiranNiranjan: This package is used because it has a ```CLI``` option which works both on windows and mac

KiranNiranjan:
@VikasShashidhar made changes. Please review. Thanks

package.json
VikasShashidhar: Ok 👍

VikasShashidhar:
LGTM 👍

PR #156, opened 2017-07-06
Electron-98 (Node Integration and Sandbox for Tests)

KiranNiranjan:
Implemented method to turn off sandbox mode when the app is running in node env.

@VikasShashidhar Please review. Thanks

### DIFF
--- a/js/notify/electron-notify.js
+++ b/js/notify/electron-notify.js
@@ -15,7 +15,7 @@ const electron = require('electron');
 const app = electron.app;
 const BrowserWindow = electron.BrowserWindow;
 const ipc = electron.ipcMain;
-const { isMac } = require('../utils/misc');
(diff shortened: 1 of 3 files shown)


js/notify/electron-notify.js
```diff
@@ -131,8 +131,8 @@ let config = {
         acceptFirstMouse: true,
         webPreferences: {
             preload: path.join(__dirname, 'electron-notify-preload.js'),
-            sandbox: true,
-            nodeIntegration: false
+            sandbox: !process.env.NODE_ENV,
```
VikasShashidhar: We can probably move this to a global variable and use that here instead.

KiranNiranjan:
@VikasShashidhar Please review. Thanks

VikasShashidhar:
LGTM 👍

PR #158, opened 2017-07-07
Electron-93 (Robotjs for AlwaysOnTop)

KiranNiranjan:
Implemented robotjs for always on top feature.

@VikasShashidhar & @VishwasShashidhar  Please review. Thanks

### DIFF
--- a/tests/alwaysOnTop.test.js
+++ b/tests/alwaysOnTop.test.js
@@ -1,5 +1,10 @@
 const Application = require('./spectron/spectronSetup');
+const {isMac} = require('../js/utils/misc.js');
+const childProcess = require('child_process');
+
 let app = new Application({});
+let robot;
+let configPath;
 
 describe('Tests for Always on top', () => {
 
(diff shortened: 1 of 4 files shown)


tests/alwaysOnTop.test.js
```diff
@@ -64,16 +98,102 @@ describe('Tests for Always on top', () => {
...
+            robot.keyTap('enter');
+            setTimeout(() => {
+                done();
+            }, 5000)
+        } else {
+            app.browserWindow.getBounds().then((bounds) => {
+                robot.setMouseDelay(200);
+                let x = bounds.x + 95;
+                let y = bounds.x + 95;
+                robot.moveMouse(x, y);
```
VikasShashidhar: Are we sure this works consistently and predictably across both Mac and Windows and when the window is resized, etc.?

KiranNiranjan: Yes. As we are calculating the position with respect to the bounds that we get from the app.

VikasShashidhar: @KiranNiranjan What about when we have the frameless window?

KiranNiranjan: Frameless window is only supported in mac and the menu bar will always be on top of the screen. so only we are handling it separately for windows.

tests/alwaysOnTop.test.js
```diff
@@ -56,24 +87,100 @@ describe('Tests for Always on top', () => {
...
+                        app.browserWindow.restore().then(() => {
+                            let x = bounds.x + 95;
+                            let y = bounds.y + 35;
+
+                            robot.moveMouseSmooth(x, y);
+                            robot.mouseClick();
+                            robot.setKeyboardDelay(200);
+                            for (let i = 0; i < 4
+                                ; i++) {
+                                robot.keyTap('down');
```
VikasShashidhar: We need to figure out a better way to handle this. What if the menu option changes in the future? For now I'll approve this, but if there is a better way, we should look for it.

tests/alwaysOnTop.test.js
```diff
@@ -56,24 +87,100 @@ describe('Tests for Always on top', () => {
...
 
-    it('should change the always on top property', () => {
-        return app.browserWindow.setAlwaysOnTop(true);
+    it('should toggle the always on top property to true', (done) => {
+        if (isMac) {
+            robot.setMouseDelay(200);
+            robot.moveMouse(190, 0);
+            robot.mouseClick();
+            for (let i = 0; i < 8; i++) {
+                robot.keyTap('down');
```
VikasShashidhar: Same as above.

PR #161, opened 2017-07-12
Electron 82: Spectron Tests for Minimize on Close

VikasShashidhar:
Have integrated RobotJS as well for both Mac and Windows.


VishwasShashidhar:
lgtm

PR #162, opened 2017-07-13
Electron-77 (User config)

KiranNiranjan:
Implemented method to update user config on installer.

@VishwasShashidhar , @VikasShashidhar & @lneir Please review. Thanks.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -138,7 +138,7 @@ function updateUserConfigMac() {
         let globalConfigPath = process.argv[2];
         let userName = process.env.USER;
 
-        childProcess.exec(`rsync -r "${globalConfigPath}" "${userConfigPath}" && chown -R "${userName}" "${userConfigPath}"`, (err) => {
+        childProcess.exec(`rsync -r "${globalConfigPath}" "${userConfigPath}" && chown -R "${userName}" "${userConfigPath}"`, {timeout: 60000}, (err) => {
             if (err) {
                 reject(err);
             }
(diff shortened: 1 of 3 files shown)


js/main.js
```diff
@@ -118,6 +131,37 @@ function setStartup(lStartup){
     });
 }
 
+// Method to overwrite user config on mac installer
```
VishwasShashidhar: Any reason we are creating a child process here in place of using node's fs api for manipulating files?

js/main.js
```diff
@@ -118,6 +131,37 @@ function setStartup(lStartup){
...
+function updateUserConfigMac() {
```
VishwasShashidhar: Any reason we are creating a child process here in place of using node's fs api for manipulating files?

js/main.js
```diff
@@ -118,6 +131,37 @@ function setStartup(lStartup){
...
+            if (err) {
+                reject(err);
+            }
+            resolve();
+        });
+    });
+}
+
+// Method to overwrite user config on windows installer
+function updateUserConfigWin() {
```
VishwasShashidhar: We should use a common method for mac and windows. Just is the isMac flag to check and take action accordingly.

js/main.js
KiranNiranjan: I tried implementing this with the `node's fs` module and the following is the issue that I faced.
 
The installer runs as a root process where the directory created will be with a different permissions so to change the permissions using the method `fs.chown(path, uid, gid, callback)` requires `uid - integer` & `gid - integer` but we don't get these values if the process is running in root.

**`Root User`**
screen shot 2017-07-14 at 8 57 33 pm

**`User`**
screen shot 2017-07-14 at 8 58 23 pm

VikasShashidhar:
@KiranNiranjan @VishwasShashidhar Is this ready?

KiranNiranjan:
This feature is working fine with the `child_process`. But there are issues using the `fs` module.

VishwasShashidhar:
In that case, we can go ahead with the child process implementation. lgtm.

PR #165, opened 2017-07-20
Electron 99

VishwasShashidhar:
1. Turns off sandboxing.
2. Removes --enable-sandbox command line launch for mac and windows by deleting the batch file and the shell script for windows and mac respectively.


installer/mac/postinstall.sh
```diff
@@ -47,9 +47,9 @@ rm -f $tempFilePath
...
-#!/bin/sh
-exec "\${0%/*}/Symphony-bin" --enable-sandbox \$@
-EOT
-chmod 755 $EXEC_PATH/Symphony
\ No newline at end of file
+chmod 755 $EXEC_PATH/Symphony
+# mv $EXEC_PATH/Symphony $EXEC_PATH/Symphony-bin
+# cat > $EXEC_PATH/Symphony << EOT
+# #!/bin/sh
+# exec "\${0%/*}/Symphony-bin" --enable-sandbox \$@
```
lneir: this shouldn't be needed anymore if removing sandbox here or are you just keeping around for later?

VishwasShashidhar: i just kept it around for later. but, you are right, it's cleaner without unwanted code, i'll push a change.

VishwasShashidhar:
@lneir @VikasShashidhar Can one of you please review the changes and merge the request?

js/notify/settings/configure-notification-position.js
```diff
@@ -27,7 +28,7 @@ let windowConfig = {
     resizable: false,
     webPreferences: {
         preload: path.join(__dirname, 'configure-notification-position-preload.js'),
-        sandbox: true,
+        sandbox: sandboxed,
```
VikasShashidhar: I would prefer this to be abstracted out. Will merge this for now to have builds ready on Jenkins. We can address this as part of refactoring.

VikasShashidhar:
LGTM 👍

PR #177, opened 2017-08-18
Search-116 (Implement Encryption and Decryption of Index)

keerthi16:
### Encryption and Decryption of Index 

- Creating a zip of the user index folder and encrypting it.
- Decrypting the encrypted file and unzipping it.

_Will be creating one more ticket for performance improvement 
currently using `zip` will be checking for `lz4` or similar_ 


serkanmulayim:
Why do we see "All checks have failed" error?

js/cryptoLib/crypto.js
```diff
@@ -0,0 +1,249 @@
...
+        throw ex;
+    }
+};
+
+/**
+ * createKeyBuffer
+ * Method which returns a buffer representing a secure key generated with PBKDF2
+ * @returns Buffer
+ */
+exports.createKeyBuffer = function() {
```
serkanmulayim: The key generated here cannot be retrieved again, because the salt is random. When you restart the application, you would lose the key. Is it possible to use the key as the userId to start with instead of randomly creating it. Because the key will come from the backend in the future. We basicly need to be able to do the encryption when a key and a folder is given. So we do not need the pbkdf2 algorithm at all.

js/cryptoLib/index.js
```diff
@@ -0,0 +1,176 @@
...
+    // TODO: Need to pass key for encryption and decryption
+    constructor() {
+
+        // will be handling after implementing in client app
+        let userId = 'user_data';
+        let INDEX_VERSION = 'v1';
+        // will be handling after implementing in client app
+
+        this.indexDataFolder = INDEX_DATA_FOLDER + '_' + userId + '_' + INDEX_VERSION;
+        this.dump = TEMPORARY_PATH;
```
serkanmulayim: what is this.dump in the production environment? I think it will be the user's config folder, right?

js/cryptoLib/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+            zipArchive.on('end', () => {
+
+                if (!fs.existsSync(`${this.dump}/content.zip`)){
+                    // will be handling after implementing in client app
+                    reject();
+                    return;
+                }
+
+                const input = fs.createReadStream(`${this.dump}/content.zip`);
```
serkanmulayim: Can we name the encrypted zip file as <permanent-index-folder-name>.zip?

js/cryptoLib/index.js
keerthi16: `userData = path.join(app.getPath('userData'));`
`TEMPORARY_PATH = isDevEnv ? path.join(__dirname, '..', '..') : userData;`

Yes, In production environment it will be user's config path.

VikasShashidhar:
@serkanmulayim That's because the Contributor License Agreement check fails. I've gotten in touch with Aaron Williamson to check regarding this. But, this shouldn't block us in any way.

js/cryptoLib/index.js
```diff
@@ -0,0 +1,177 @@
...
+const path = require('path');
+const fs = require('fs');
+const archiver = require('archiver');
+const zipArchive = archiver('zip');
+const extract = require('extract-zip');
+const isDevEnv = require('../utils/misc.js').isDevEnv;
+const crypto = require('./crypto');
+
+const userData = path.join(app.getPath('userData'));
+const INDEX_DATA_FOLDER = isDevEnv ? './data/search_index' : path.join(userData, 'data/search_index');
```
VikasShashidhar: Can't we not hardcode the path here and see if it can be done in any other way?

package.json
```diff
@@ -89,17 +89,18 @@
...
     "electron-dl": "^1.9.0",
     "electron-spellchecker": "^1.2.0",
     "electron-squirrel-startup": "^1.0.0",
+    "extract-zip": "^1.6.5",
     "filesize": "^3.5.10",
     "keymirror": "0.1.1",
-    "winreg": "^1.2.3",
-    "archiver": "latest",
-    "extract-zip": "latest"
+    "node-forge": "^0.7.1",
```
lneir: forge lib already exists in sfe-client code... why re-add here?  will end up having two copies of forge!  should consider re-using that existing lib.

keerthi16: @lneir 
This is for temporary purpose only. Will be retrieving the key from the backend
and will be removing the `node-forge` package once the backend APIs are ready.

js/cryptoLib/index.js
keerthi16: I'll take care of this during refactoring.

VikasShashidhar: 👍

package.json
VikasShashidhar: 👍

VikasShashidhar:
@keerthi16 There are merge conflicts. Can you please resolve them?

package.json
keerthi16: I have created [SEARCH-206](https://perzoinc.atlassian.net/browse/SEARCH-206) to address this.

PR #178, opened 2017-08-21
Electron-114 (Reload Context Menu)

KiranNiranjan:
Added a condition to not include reload in context menu when clicked on links.


VishwasShashidhar:
fyi, reload with a right click on a link doesn't work (reason to be found out why yet), this is the case with Slack as well, hence, we've disabled the "reload" button when someone right clicks on a link in the chat window.

VikasShashidhar:
LGTM 👍

PR #179, opened 2017-08-22
Electron 113 (User Config Changes)

VishwasShashidhar:
JIRA Ticket Reference -> https://perzoinc.atlassian.net/browse/ELECTRON-113

Following changes have been implemented as part of this PR

- for single user install: - over write user config with new values from install. - put same info into global config. - during uninstall remove global and user config
- for multi user install: - during install don't touch user config - put install parameters into global config - during uninstall don't touch user config and remove global config.
- in both cases, we'll use the same logic that user config value over-rides global config value.
on mac: installing into applications should be same as multi-user install. installing into some private location should be the same as single user install.
- also when updating user config we should preserve other values, like notificationSettings and window locations. that is, we should only be updating parameters from installer (url, minimize on close, launch on startup and always on top.


VishwasShashidhar:
@lneir @VikasShashidhar Please review and merge.

CC: @KiranNiranjan

package.json
```diff
@@ -1,108 +1,112 @@
...
-      "!node_modules/@paulcbetts/cld/deps/cld${/*}",
-      "!node_modules/@paulcbetts/cld/build/deps${/*}",
-      "!node_modules/@paulcbetts/spellchecker/vendor${/*}"
-    ],
-    "extraFiles": "config/Symphony.config",
-    "appId": "symphony-electron-desktop",
-    "mac": {
-      "target": "dmg",
-      "category": "public.app-category.business"
+    "name": "Symphony",
```
lneir: is anything changing here - or just spacing?

js/main.js
```diff
@@ -123,8 +120,8 @@ function setupThenOpenMainWindow() {
         // as the app is launched as a root user we don't get
         // access to the config file
         let launchOnStartup = process.argv[3];
-        updateUserConfigMac()
-            .then(setStartup(launchOnStartup))
+        setStartup(launchOnStartup)
+            .then(() => updateUserConfigMac(process.argv[2]))
```
lneir: what is argv[2]?  appears to be global path.  is the reliable?

js/config.js
```diff
@@ -182,6 +187,159 @@ function saveUserConfig(fieldName, newValue, oldConfig) {
...
+        });
+    });
+}
+
+/**
+ * Method to read user config data
+ * @param {String} userConfigFile - The user config file path
+ * @returns {Promise}
+ */
+function getUserConfigData(userConfigFile) {
```
lneir: there is already a function above 'readUserConfig' that does a lot of this - seems like duplication - can we collapse into one func?

js/config.js
```diff
@@ -182,6 +187,159 @@ function saveUserConfig(fieldName, newValue, oldConfig) {
...
+function updateUserConfig(newGlobalConfig, oldUserConfig) {
+    return new Promise((resolve, reject) => {
+        // Picking some values from global config to overwrite user config
+        const configDataToUpdate = pick(newGlobalConfig, ['url', 'minimizeOnClose', 'launchOnStartup', 'alwaysOnTop']);
+        const updatedUserConfigData = Object.assign(oldUserConfig, configDataToUpdate);
+        const jsonNewConfig = JSON.stringify(updatedUserConfigData, null, ' ');
+        // get user config path
+        let userConfigFile;
+
+        if (isMac) {
```
lneir: why something special for user config update - saveUserConfig doesn't have to do anything special?

js/config.js
```diff
@@ -182,6 +187,159 @@ function saveUserConfig(fieldName, newValue, oldConfig) {
...
+        });
+    });
+}
+
+/**
+ * Method which copies global config file to user config directory for mac
+ * @param {String} globalConfigPath - The global config path from installer
+ * @returns {Promise}
+ */
+function copyConfigMac(globalConfigPath) {
```
lneir: why do we need to pass globalConfigPath here - shouldn't it be relative to exe like code in readGlobalConfig?

lneir:
as per electron-113 ticket: for single user install: during uninstall should remove global and user config - i don't see update to do this???

package.json
VishwasShashidhar: It is just spacing.

js/main.js
VishwasShashidhar: With the post install script, we pass in an argument to get the global Symphony.config path file (i.e. /Applications/Symphony.app/Contents/config/Symphony.config). We unfortunately don't get this directly from the app as it'll be run from a temporary path and any relative path for the config file doesn't work and we end up manipulating the incorrect config file.

Regardless, I've abstracted this to a variable for better clarity.

js/config.js
VishwasShashidhar: This method is to update multiple fields at once. saveUserConfig allows us to update only one field at once.

js/config.js
VishwasShashidhar: will fix it.

js/config.js
VishwasShashidhar: With the post install script, we pass in an argument to get the global Symphony.config path file (i.e. /Applications/Symphony.app/Contents/config/Symphony.config). We unfortunately don't get this directly from the app as it'll be run from a temporary path and any relative path for the config file doesn't work and we end up manipulating the incorrect config file. That's the reason to pass this as an argument.

VishwasShashidhar:
it's tricky to capture the uninstall event. Currently, squirrel supports it, but, we don't use the squirrel installer in the first place as we use the win-unpacked directory with aip.

we might have to see if it is possible via aip and process the event there.

lneir:
if we don't support removing the user symphony.config for single install.  then user might install for single user and later install for all users then you'll have url in user config (from single install) and it won't properly read the url from the install for all.  or maybe we need to rethink how this should work.

VishwasShashidhar:
can we take the uninstall part in a separate ticket and merge these changes?

VikasShashidhar:
@lneir Waiting on your clearance for this.

VikasShashidhar:
@VishwasShashidhar Can you please resolve merge conflicts on this?
@lneir Should we take up the uninstallation part as a separate ticket and merge this?

VishwasShashidhar:
@VikasShashidhar Done, merged master to this branch and resolved conflicts.

PR #181, opened 2017-08-28
Search-154 & Search-116 (Changed encryption implementation)

keerthi16:
### Changed encryption implementation

- Passing the key and the userId from the client app
- Creating a encrypted file as soon as the batch indexes are merged
- Deleting the data folder when the closes

### DIFF
--- a/js/cryptoLib/index.js
+++ b/js/cryptoLib/index.js
@@ -27,7 +27,6 @@ class Crypto {
         this.permanentIndexFolderName = 'search_index_' + userId + '_' + INDEX_VERSION;
         this.dump = TEMPORARY_PATH;
         this.extractToPath = `${TEMPORARY_PATH}/data/${this.permanentIndexFolderName}`;
-        this.key = "XrwVgWR4czB1a9scwvgRUNbXiN3W0oWq7oUBenyq7bo="; // temporary only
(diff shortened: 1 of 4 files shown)


js/cryptoLib/index.js
```diff
@@ -142,36 +132,37 @@ class Crypto {
...
             }
         });
     }
 
     /**
+     * Deleting the data index folder
+     * when the app is closed
+     */
+    deleteFolders() {
+        Crypto.deleteFolderRecursive(this.dataFolder);
```
VikasShashidhar: Why do we need to delete folders recursively?

My assumption is -> While closing the app, we would only have one user's directory in the data folder, correct?

js/search/search.js
```diff
@@ -45,17 +51,27 @@ class Search {
...
-        this.init();
+        this.crypto = new Crypto(userId, key);
+        this.decryptAndInit();
+    }
+
+    decryptAndInit() {
+        this.crypto.decryption().then(() => {
+            this.init();
+        }).catch(() => {
+            this.init();
```
VikasShashidhar: What's exactly going on here?

keerthi16: Ensuring the search library is initialised even if the decryption fails

js/cryptoLib/index.js
keerthi16: We cannot delete a directory that is not empty, So to remove a dir we need to recursively remove all the files and folder inside it.

PR #186, opened 2017-09-07
Electron-118 (Screen snippet always on top)

KiranNiranjan:
Implemented a method to turn off `alwaysOnTop` while using screen snippet (windows).

@VikasShashidhar , @VishwasShashidhar & @lneir Please review. Thanks.

### DIFF
--- a/js/screenSnippet/index.js
+++ b/js/screenSnippet/index.js
@@ -10,10 +10,12 @@ const path = require('path');
 const { isMac, isDevEnv } = require('../utils/misc.js');
 const log = require('../log.js');
 const logLevels = require('../enums/logLevels.js');
+const eventEmitter = require('.././eventEmitter');
 
 // static ref to child process, only allow one screen snippet at time, so
 // hold ref to prev, so can kill before starting next snippet.
 let child;
+let isAlwaysOnTop;
 
 /**
  * Captures a user selected portion of the monitor and returns jpeg image
@@ -61,6 +63,17 @@ class ScreenSnippet {
                     captureUtil = path.join(execPath, 'ScreenSnippet.exe');
                 }
 
+                // Method to verify and disable always on top property
+                // as an issue with the ScreenSnippet.exe not being on top
+                // of the electron wrapper
+                const windows = electron.BrowserWindow.getAllWindows();
+                if (windows && windows.length > 0) {
+                    isAlwaysOnTop = windows[ 0 ].isAlwaysOnTop();
+                    if (isAlwaysOnTop) {
+                        eventEmitter.emit('isAlwaysOnTop', false);
+                    }
+                }
+
                 captureUtilArgs = [outputFileName];
             }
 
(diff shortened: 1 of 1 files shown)


js/screenSnippet/index.js
```diff
@@ -61,6 +63,17 @@ class ScreenSnippet {
                     captureUtil = path.join(execPath, 'ScreenSnippet.exe');
                 }
 
+                // Method to verify and disable always on top property
+                // as an issue with the ScreenSnippet.exe not being on top
+                // of the electron wrapper
+                const windows = electron.BrowserWindow.getAllWindows();
+                if (windows && windows.length > 0) {
+                    isAlwaysOnTop = windows[ 0 ].isAlwaysOnTop();
```
KiranNiranjan: Is it fine if i read this from electron api or should i read it from config file?

@VikasShashidhar, @VishwasShashidhar & @lneir

VishwasShashidhar: Either way should be fine. I'd prefer using the electron api as it is dynamic.

js/screenSnippet/index.js
```diff
@@ -74,9 +87,17 @@ class ScreenSnippet {
...
                 if (error && error.killed) {
+                    // Method to reset always on top feature
+                    if (isAlwaysOnTop) {
+                        eventEmitter.emit('isAlwaysOnTop', true);
+                    }
                     // processs was killed, just resolve with no data.
                     resolve();
                 } else {
+                    // Method to reset always on top feature
+                    if (isAlwaysOnTop) {
```
VishwasShashidhar: Can we eliminate duplicate code for the if condition and move it after the `if-else` block?

js/screenSnippet/index.js
```diff
@@ -61,6 +63,17 @@ class ScreenSnippet {
                     captureUtil = path.join(execPath, 'ScreenSnippet.exe');
                 }
 
+                // Method to verify and disable always on top property
+                // as an issue with the ScreenSnippet.exe not being on top
+                // of the electron wrapper
+                const windows = electron.BrowserWindow.getAllWindows();
```
lneir: this looks like it is only is changing isAlwaysOnTop value for first windows (which i would assume is main window) but have you tried this for the pop-outs.  i would assume same issue will happen there.

KiranNiranjan: I'm just reading the value here. If `alwaysOnTop` is enabled i'm using the `eventEmitter` to `setAlwaysOnTop` for all the window in `windowMgr.js:501`

lneir: yes but main question: what happens when always on top is enable and you take screensnippet from a pop-out.  does the issue happen there as well?  or this only an issue for the main window?

KiranNiranjan: Yes this issue happens both on pop-out and main window.

lneir: ok i see, lgtm.

PR #191, opened 2017-09-26
Add tooltip to display full filename in download manager.

VikasShashidhar:
Added a browser tooltip to display full filename in download manager. This mimics the default behaviour of the Chrome Download Manager.

@lneir @VishwasShashidhar Please review.


VishwasShashidhar:
Looks good. But, can we write a unit test for this?

PR #192, opened 2017-09-26
Adds a count to the file name in the download manager if the user dow…

VikasShashidhar:
Adds a count to the file name in the download manager if the user downloads a file with the same name.

This mimics the Google Chrome behaviour.

@lneir @VishwasShashidhar Please review.


js/downloadManager/index.js
```diff
@@ -194,4 +195,33 @@ function initiate() {
             });
         }
     }
+}
+
+/**
+ * Return a file display name for the download item
+ */
+function getFileDisplayName(fileName) {
```
lneir: what about case where i download file.  
then refresh client.
now download again.
 
the download filename should reflect the actual version in the downloads folder - that is what chrome is doing.

VikasShashidhar: Chrome persists all the data (file name, URL, etc.) that we download and makes it available to us. Currently, both the Paragon and the Electron Wrapper don't have such an option.

VikasShashidhar: Waiting for input from Product -> https://perzoinc.atlassian.net/browse/ELECTRON-146

VikasShashidhar: @lneir After discussions with Thibaud last week, can we consider this as a minor issue and merge this PR? We already have an open ticket to track the persistent behaviour of download bar.

PR #200, opened 2017-10-06
Electron 151: Fixes link crash issue

VishwasShashidhar:
This fixes an issue with electron crashes on clicking links continuously. We use electron 1.7.8 here.

@VikasShashidhar Please review and merge


package.json
```diff
@@ -76,7 +76,7 @@
   "devDependencies": {
     "browserify": "^14.1.0",
     "cross-env": "^3.2.4",
-    "electron": "1.7.5",
+    "electron": "^1.7.8",
```
lneir: i see another pr upgrading to 1.8.1???https://github.com/symphonyoss/SymphonyElectron/pull/204/files

VishwasShashidhar: Yes, but, 1.8.1 is still a beta and they've mentioned that there may be regressions due to upgrades on node, chrome and v8. So, using 1.7.8 to fix this critical issue.

We can keep the other PR open until we have a stable 1.8.1 from electron.

PR #201, opened 2017-10-06
Electron 135: Pop out window -> url opens in external browser

VishwasShashidhar:
**Problem**

When we try to open an external link from inside a pop out window, it opens the link inside electron creating a new window which isn't the expected behavior. Instead, an external link should always be opened in the default browser of the OS.

**Solution Approach**

On a pop-out (child window), we listen for the "new-window" event and prevent the default event from occurring and instead open the link in an external browser.

@VikasShashidhar Please review and merge.

CC: @lneir

### DIFF
--- a/js/windowMgr.js
+++ b/js/windowMgr.js
@@ -331,13 +331,22 @@ function doCreateMainWindow(initialUrl, initialBounds) {
                     browserWin.winName = frameName;
                     browserWin.setAlwaysOnTop(alwaysOnTop);
 
-                    browserWin.once('closed', function () {
+                    let handleChildWindowClosed = () => {
                         removeWindowKey(newWinKey);
                         browserWin.removeListener('move', throttledBoundsChange);
-                        browserWin.removeListener('resize', throttledBoundsChange);
+                        browserWin.removeListener('resize', throttledBoundsChange);                        
+                    };
+
+                    browserWin.once('closed', () => {
+                        handleChildWindowClosed();
                     });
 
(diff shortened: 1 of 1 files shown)


js/windowMgr.js
```diff
@@ -365,6 +355,14 @@ function doCreateMainWindow(initialUrl, initialBounds) {
                         });
                     });
 
+                    // In case we navigate to an external link from inside a pop-out,
+                    // we open that link in an external browser rather than creating
+                    // a new window
+                    browserWin.webContents.on('new-window', (childEvent, childWinUrl) => {
```
lneir: make sure event is cleaned up when pop-out is closed.
also same above for 'crashed' event
or at least ensure no leakage occurs.

VishwasShashidhar: done.

js/windowMgr.js
```diff
@@ -357,14 +347,28 @@ function doCreateMainWindow(initialUrl, initialBounds) {
...
+                        childEvent.preventDefault();
+                        openUrlInDefaultBrowser(childWinUrl);
+                    });
+
+                    // Clear up the browser window once the window is closed
+                    // to avoid leakage
+                    browserWin.on('closed', () => {
```
lneir: there is already a close above on line 334.  there you can remove event handler.
how will setting browserWin to null clean up?

VishwasShashidhar:
@lneir I've made the necessary changes, please review.

PR #202, opened 2017-10-06
Electron-171 (Configuration window positioning)

KiranNiranjan:
Fixes the configuration window positioning when multiple monitors are connected

### DIFF
--- a/js/notify/settings/configure-notification-position.js
+++ b/js/notify/settings/configure-notification-position.js
@@ -87,7 +87,7 @@ function openConfigurationWindow(windowName) {
         const windowWidth = Math.round(width * 0.5);
         const windowHeight = Math.round(height * 0.5);
 
-        // Calculating the center of the window
+        // Calculating the center of the parent window
         // to place the configuration window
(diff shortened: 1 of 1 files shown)


js/notify/settings/configure-notification-position.js
```diff
@@ -72,13 +72,27 @@ function getTemplatePath() {
...
 
     // if we couldn't find any window matching the window name
     // it will render as a new window
-    if (allWindows) {
-        windowConfig.parent = allWindows;
+    if (selectedParentWindow) {
+        windowConfig.parent = selectedParentWindow;
+
+        // This is a temporary work around until there
+        // is a fix for the modal window in windows from the electron
```
lneir: this there an electron bug for this?  if so, put here.

KiranNiranjan: Electron issue [here](https://github.com/electron/electron/issues/10721)

lneir: i would add that to documentation in code.

PR #204, opened 2017-10-06
electron-155: fixes the raised issue

VishwasShashidhar:
**Problem**

When we have a pop out window open and close the main window, an exception is thrown by electron as per this screenshot. screen shot 2017-10-06 at 16 57 57

**Solution Approach**

This is a known electron issue and has been fixed as part of electron 1.8.1. For more details, please check the JIRA ticket.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   "devDependencies": {
     "browserify": "^14.1.0",
     "cross-env": "^3.2.4",
-    "electron": "1.7.5",
+    "electron": "^1.8.1",
     "electron-builder": "^13.9.0",
     "electron-builder-squirrel-windows": "^12.3.0",
     "electron-packager": "^8.5.2",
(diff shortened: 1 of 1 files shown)


VishwasShashidhar:

package.json
```diff
@@ -76,7 +76,7 @@
   "devDependencies": {
     "browserify": "^14.1.0",
     "cross-env": "^3.2.4",
-    "electron": "1.7.5",
+    "electron": "^1.8.1",
```
lneir: looks good but i would used a fix version: 1.8.1

VishwasShashidhar: Makes sense. I've pushed an update.

I tried building with node 6.10.0, didn't see any problems. I guess we should be good to go.

VikasShashidhar:
I'll keep this PR open for now. Let's wait for a stable 1.8.1. This is not a blocker.

VishwasShashidhar:
will close this as it is confusing. Will raise a new PR once we have the stable version of 1.8.1 of electron.

PR #207, opened 2017-10-09
Electron 141: fix on url validation issue on installer

VishwasShashidhar:
**Problem**

During the installation, there is no validation currently done on the pod url which is leading to bad installations leading to users having to manually change the pod url in the global config file.

**Solution Approach**

On mac, we have events to listen on and add validations and let the user continue / not continue to the next screen in the InstallerPanel. And, that is implemented as per the below screenshot. 
screen shot 2017-10-09 at 19 32 12

On windows, we don't have any such events to listen on a text box, so, we've had to use the "Next" button to perform the validation and show another screen for verification purpose and perform the validation there by disabling the "Install" button as per the below screenshot. 
screen shot 2017-10-02 at 21 54 48
screen shot 2017-10-02 at 21 58 23


VishwasShashidhar:

VikasShashidhar:
LGTM 👍

installer/win/SymphonyWinInstallerValidatorAction/SymphonyWinInstallerValidatorAction/CustomAction.cs
```diff
@@ -0,0 +1,102 @@
+using System;
```
lneir: will this only require .net3.5 or higher.  most customer will have upgraded to newer version of .net but out of the box win7 only has .net3.5.

VishwasShashidhar: you are right, this isn't working on Win7. I'll need to find a soln. for it.

VishwasShashidhar:
@lneir I've removed the dll dependency as I was finally able to get it working with JScript in AIP.

This should be good to go now. Please review.

CC: @VikasShashidhar @KiranNiranjan

installer/mac/SymphonySettingsPlugin/SymphonySettingsPlugin/MyInstallerPane.m
```diff
@@ -14,21 +14,40 @@ - (NSString *)title
...
-        podUrl = @"my.symphony.com";
+    // Check if the url contains a protocol, if not, prepend https to it
+    NSString *prefix = @"https://";
+    if (![podUrl hasPrefix:prefix]) {
+        podUrl = [prefix stringByAppendingString:podUrl];
+        [_podUrlTextBox setStringValue:podUrl];
+    }
+    
+    // Now, validate the url against a url regex
+    NSString *regex = @"^((?:http:\/\/)|(?:https:\/\/))(www.)?((?:[a-zA-Z0-9]+\.[a-z]{3})|(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(?::\d+)?))([\/a-zA-Z0-9\.]*)$";
```
VikasShashidhar: Having only "www" might be a problem here. What if PODs are run on subdomains?

VishwasShashidhar: `www` is an optional (`?` indicates optional). This works for sub-domains too. Please check https://regex101.com/ with the above regex, both sub-domains and full domains work with or without `www`.

PR #210, opened 2017-10-09
Electron-170 (Monitor position in notification settings)

KiranNiranjan:
Added an event to update the monitor selection drop-down whenever a display is added/removed.


VishwasShashidhar:
@KiranNiranjan Can you add a code comment explaining what the code does?

PR #211, opened 2017-10-10
Electron-134 (Notification redesign)

KiranNiranjan:
Redesigned the notifications for windows OS.

notification


js/notify/electron-notify.html
```diff
@@ -2,11 +2,18 @@
...
 <div id="container">
-	<div id="header">
-		<img src="" id="image" />
+    <div style="overflow: hidden; position: absolute; transform: translate(-25px, 10px);">
+        <img src="" id="symphony-logo">
+    </div>
+    <div id="header">
 		<span id="title"></span>
+        <!-- Hard coding this until the backend is ready -->
+        <span id="pod">symphony</span>
```
lneir: see comment in jira ticket.  should we show at all if we don't have correct pod info.  check with product and design.

js/notify/electron-notify-preload.js
```diff
@@ -75,6 +78,22 @@ function setContents(event, notificationObj) {
 
     if (notificationObj.color) {
         container.style.backgroundColor = notificationObj.color;
+
+        if (notificationObj.color.match(/^(?:white|#fff(?:fff)?|rgba?\(\s*255\s*,\s*255\s*,\s*255\s*(?:,\s*1\s*)?\))$/i)) {
```
VikasShashidhar: Better to abstract this out. If in the future there is another lighted color (yellow, biege, etc.), we may have to use the black logo.

js/notify/electron-notify-preload.js
```diff
@@ -75,6 +78,22 @@ function setContents(event, notificationObj) {
...
+            let logo = notiDoc.getElementById('symphony-logo');
+            logo.src = './assets/symphony-logo-black.png';
+        } else {
+
+            let logo = notiDoc.getElementById('symphony-logo');
```
VikasShashidhar: Can be put outside this condition and remove the definition in the "if" block.

js/notify/electron-notify.html
```diff
@@ -2,11 +2,17 @@
 <head></head>
 <body style='margin:0; overflow: hidden; -webkit-user-select: none;'>
 <div id="container">
-	<div id="header">
-		<img src="" id="image" />
+    <div style="overflow: hidden; position: absolute; transform: translate(-25px, 10px);">
```
VikasShashidhar: Can you avoid using absolute positioning?

js/notify/electron-notify.js
```diff
@@ -76,17 +76,16 @@ let config = {
...
     },
     defaultStyleImage: {
-        flex: '0 0 auto',
-        overflow: 'hidden',
-        height: 30,
-        width: 30,
-        marginLeft: 0,
-        marginRight: 8
+        height: 43,
+        position: 'absolute',
```
VikasShashidhar: Same as above

js/notify/electron-notify.js
```diff
@@ -99,28 +98,37 @@ let config = {
         color: '#CCC'
     },
     defaultStyleTitle: {
-        fontFamily: 'Arial',
+        fontFamily: 'Lato-Bold, san-serif',
```
VikasShashidhar: Typo - it's "sans-serif".

js/notify/electron-notify.js
```diff
@@ -99,28 +98,37 @@ let config = {
...
-        alignSelf: 'center',
+        color: '#4a4a4a',
+        marginLeft: 52,
+        overflow: 'hidden',
+        display: '-webkit-box',
+        webkitLineClamp: 1,
+        webkitBoxOrient: 'vertical',
+    },
+    defaultStylePod: {
+        fontFamily: 'Lato-Regular, san-serif',
```
VikasShashidhar: Same as above. Also, I wonder why we should use Lato here. Are we including the font as part of the package or in any other way? If not, it's not really useful as Lato is never almost included as part of OS fonts.

js/notify/electron-notify.js
```diff
@@ -99,28 +98,37 @@ let config = {
...
     },
     defaultStyleText: {
-        flex: '0 0 auto',
-        fontFamily: 'Calibri',
-        fontSize: 14,
-        fontWeight: 400,
-        opacity: 0.87,
-        margin: 0,
-        marginTop: 4,
+        fontFamily: 'Lato-Regular, san-serif',
```
VikasShashidhar: Same as above.

KiranNiranjan:
@VikasShashidhar Made those changes. Please review. Thanks

PR #213, opened 2017-10-11
Electron-175 (AlwaysOnTop for about Symphony window)

KiranNiranjan:
Method to set the AlwaysOnTop property for about window.

@VikasShashidhar , @VishwasShashidhar & @lneir Please Review. Thanks

### DIFF
--- a/js/aboutApp/index.js
+++ b/js/aboutApp/index.js
@@ -67,6 +67,13 @@ function openAboutWindow(windowName) {
     aboutWindow.setVisibleOnAllWorkspaces(true);
     aboutWindow.loadURL(getTemplatePath());
 
+    // sets the AlwaysOnTop property for the about window
+    // if the main window's AlwaysOnTop is true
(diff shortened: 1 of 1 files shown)


js/aboutApp/index.js
```diff
@@ -67,6 +67,13 @@ function openAboutWindow(windowName) {
     aboutWindow.setVisibleOnAllWorkspaces(true);
     aboutWindow.loadURL(getTemplatePath());
 
+    // sets the AlwaysOnTop property for the about window
+    // if the main window's AlwaysOnTop is true
+    let focusedWindow = BrowserWindow.getFocusedWindow();
+    if (focusedWindow && focusedWindow.isAlwaysOnTop()) {
+        aboutWindow.setAlwaysOnTop(true);
```
VikasShashidhar: We may not even need this as the Windows version may lose the menu bar. I'll keep this open for now.

PR #214, opened 2017-10-16
Electron-154 (Menu items issue on windows)

KiranNiranjan:
Optimized reading config fields which fix the menu items issue on Windows.

## Approach

- **Problem with the code:** There were multiple requests sent to read the config for every single field and the menu template was returned even before they returned promise.

- **Fix:** Modified the method to read and parse multiple fields so that single request can be made to read all the config fields

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -6,6 +6,8 @@ const path = require('path');
 const fs = require('fs');
 const AppDirectory = require('appdirectory');
 const omit = require('lodash.omit');
+const pick = require('lodash.pick');
+const difference = require('lodash.difference');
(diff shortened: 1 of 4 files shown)


VishwasShashidhar:
Is the method `setupThenOpenMainWindow` still in the code? If so, wouldn't it be better to remove it since we'll not be using it?

KiranNiranjan:
@VishwasShashidhar we are using the `setupThenOpenMainWindow` method after we read the data from the config file.

VishwasShashidhar:
Ok

Issue #205, opened 2017-10-06
npm install fails - error in spellchecker_mac.mm code


bluechips23:
I did a clone of the master branch this morning and I ran `npm install` and I got the following error. Seems like there's an error in the code but let me know if I'm doing something wrong.

I'm running on MacOS - 10.12.6 (Sierra).

Error:

  ```
LIBTOOL-STATIC Release/hunspell.a
  CXX(target) Release/obj.target/spellchecker/src/main.o
  CXX(target) Release/obj.target/spellchecker/src/spellchecker_hunspell.o
  **CXX(target) Release/obj.target/spellchecker/src/spellchecker_mac.o
../src/spellchecker_mac.mm:184:13: error: cannot initialize a variable of type 'NSString *' with an rvalue of type 'const NSString *'
  NSString* globalLang = currentGlobalLanguage ? currentGlobalLanguage : autoLangauge;
            ^            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/spellchecker_mac.mm:185:13: error: cannot initialize a variable of type 'NSString *' with an rvalue of type 'const NSString *'
  NSString* ourLang = this->spellCheckerLanguage ? this->spellCheckerLanguage : autoLangauge;
            ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.**
make: *** [Release/obj.target/spellchecker/src/spellchecker_mac.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:258:23)
gyp ERR! stack     at emitTwo (events.js:125:13)
gyp ERR! stack     at ChildProcess.emit (events.js:213:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:200:12)
gyp ERR! System Darwin 16.7.0
gyp ERR! command "/usr/local/Cellar/node/8.6.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
[...]

PR #215, opened 2017-10-17
removed package-lock.json from git cache and added it to gitignore

VishwasShashidhar:
**Problem**

With package-lock.json baked into the repo, if we've made any updates to the package.json file, users running on npm > 5 is experiencing an issue (not really a bug, but a behaviour problem) where in the dependencies are being pulled from the package-lock.json with package.json deemed useless.

**Solution**

Since we are primarily using package.json while creating builds on Jenkins, I've removed the package-lock.json from git cache and added it to the ignore list.

Once we establish a process of using package-lock.json and npm > 5, we can add it back to the git list.

PS: An alternative is -> https://codeburst.io/disabling-package-lock-json-6be662f5b97d


VishwasShashidhar:

CC: @pmonks @bluechips23

VishwasShashidhar:
Observed on https://github.com/symphonyoss/SymphonyElectron/issues/205#issuecomment-337115372

pmonks:
Manually deleting `package-lock.json` and rebuilding got me past this issue, so this LGTM!

PR #217, opened 2017-10-18
Electron-174 (Basic authentication)

KiranNiranjan:
Added a support for electron basic authentication.

## Windows
windows

## macOS
screen shot 2017-10-18 at 12 47 01 pm


js/basicAuth/basic-auth.html
```diff
@@ -0,0 +1,98 @@
...
+            text-align: start;
+        }
+
+        .hostname {
+            font-size: .9em;
+        }
+
+        label {
+            display: inline-block;
+            width: 20%;
```
VikasShashidhar: Why set a percentage width here when we have pixel dimensions for the window?

PR #222, opened 2017-10-30
Electron-142 (Build Number)

KiranNiranjan:
Included build version in the about app window.

screen shot 2017-10-30 at 2 28 49 pm

### DIFF
--- a/js/aboutApp/renderer.js
+++ b/js/aboutApp/renderer.js
@@ -22,6 +22,6 @@ ipcRenderer.on('buildVersion', (event, buildVersion) => {
     const version = remote.app.getVersion();
 
     if (versionText) {
-        versionText.innerHTML = version ? `Version ${version} (${version}.${buildVersion})` : null;
+        versionText.innerHTML = version ? `Version ${version} (${version}.${buildVersion})` : 'N/A';
     }
 });
--- a/js/aboutApp/renderer.js
+++ b/js/aboutApp/renderer.js
@@ -1,5 +1,5 @@
 'use strict';
-const { remote } = require('electron');
+const { remote, ipcRenderer } = require('electron');
 
 renderDom();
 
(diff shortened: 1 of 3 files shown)


js/aboutApp/renderer.js
```diff
@@ -9,13 +9,19 @@ renderDom();
...
         copyright.innerHTML = `Copyright &copy; ${new Date().getFullYear()} ${applicationName}`
     });
 }
+
+ipcRenderer.on('buildVersion', (event, buildVersion) => {
+    let versionText = document.getElementById('version');
+    const version = remote.app.getVersion();
+
+    if (versionText) {
+        versionText.innerHTML = version ? `Version ${version} (${version}.${buildVersion})` : null;
```
VishwasShashidhar: Instead of `null`, let's make this `N/A`

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "Symphony",
   "productName": "Symphony",
   "version": "2.0.0",
+  "buildVersion": "",
```
VishwasShashidhar: Can we also call this "buildNumber" in place of "buildVersion".

VishwasShashidhar:
lgtm

js/aboutApp/renderer.js
```diff
@@ -9,13 +9,19 @@ renderDom();
...
         copyright.innerHTML = `Copyright &copy; ${new Date().getFullYear()} ${applicationName}`
     });
 }
+
+ipcRenderer.on('buildNumber', (event, buildNumber) => {
+    let versionText = document.getElementById('version');
+    const version = remote.app.getVersion();
+
+    if (versionText) {
+        versionText.innerHTML = version ? `Version ${version} (${version}.${buildNumber})` : 'N/A';
```
VikasShashidhar: Does this work for Mac and Windows?

KiranNiranjan: This feature is only for windows as we are restricting it in the menu template here: https://github.com/symphonyoss/SymphonyElectron/blob/8fb57d23759da6582c81ef0f3f362fdace185ec9/js/menus/menuTemplate.js#L243

PR #226, opened 2017-11-07
Electron 201: auto launch fixes

VishwasShashidhar:
**Issue**
An issue was reported by a sales representative who had trouble with auto launch. Issue was that Symphony was being launched even if the setting was turned off during installation.

More details here -> https://perzoinc.atlassian.net/browse/ELECTRON-201

**Problem**
During installation, we invoke a method called `setStartup` to manage the auto launch settings, this method doesn't handle the autoLaunch argument sent from the installation in the necessary manner to convert the argument string to boolean.

**Solution**
Converting the argument string to boolean fixes the issue.


VishwasShashidhar:

CC: @KiranNiranjan

PR #227, opened 2017-11-07
Electron-197 (Windows Desktop composition)

KiranNiranjan:
Added a check to only return screen sources if desktop composition is disabled on Windows


js/desktopCapturer/getSources.js
```diff
@@ -52,6 +53,17 @@ function getSources(options, callback) {
         };
     }
 
+    if (!isMac) {
```
VikasShashidhar: I'd rather you check if this is Windows instead of not Mac. The below code will break if it's anything apart from Windows.

PR #232, opened 2017-11-10
Electron 194: Enables setting chrome flags

VishwasShashidhar:
**Issue**
As part of [electron-194](https://perzoinc.atlassian.net/browse/ELECTRON-194), we need to support setting chrome flags for auth whitelist.

**Approach**
Currently, an admin will have to make changes in the global Symphony.config and set the parameters "authServerWhitelist" and "authNegotiateDelegateWhitelist" under the "customFlags" root parameter.

We read these parameters while initializing main.js and set it via a command line switch in electron and to the chrome flags.

**Caveats**
- There's no UI to set these parameters currently
- After making any changes to the global Symphony.config file, an user will have to restart the Symphony Electron app if it is already running for the changes to take effect.


VishwasShashidhar:

PR #237, opened 2017-11-14
Electron-211 (Added company filed to the notifications)

KiranNiranjan:
Updated the toast notification to display company information.

company

### DIFF
--- a/js/notify/electron-notify-preload.js
+++ b/js/notify/electron-notify-preload.js
@@ -23,7 +23,7 @@ function setStyle(config) {
     let image = notiDoc.getElementById('image');
     let logo = notiDoc.getElementById('symphony-logo');
     let title = notiDoc.getElementById('title');
-    let pod = notiDoc.getElementById('pod');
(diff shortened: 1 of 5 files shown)


js/notify/electron-notify-preload.js
```diff
@@ -88,13 +94,9 @@ function setContents(event, notificationObj) {
...
         } else {
-            let title = notiDoc.getElementById('title');
-            let pod = notiDoc.getElementById('pod');
-            let message = notiDoc.getElementById('message');
-
-            message.style.color = '#ffffff';
-            title.style.color = '#ffffff';
-            pod.style.color = notificationObj.color;
+            messageDoc.style.color = '#ffffff';
+            titleDoc.style.color = '#ffffff';
```
VikasShashidhar: What about in cases where we have a light theme?

KiranNiranjan: Other than white there are no light colours defined in the client app for the notifications.

We don't need to modify anything until they define colours lighter than these 
colors

PR #238, opened 2017-11-16
Electron 205 - Allow user to set Downloads folder through Menu dropdown

VishwasShashidhar:
[**Issue**](https://perzoinc.atlassian.net/browse/ELECTRON-205)

Currently, with the electron wrapper, we don't have an option to download files onto a path that we want to. All the files are downloaded to the default download directory of the underlying OS.

This issue addresses the above problem by supporting a user to set the downloads directory specifically for electron.

**Approach**

We would be able to set the downloads directory through a menu option which opens a dialog through which we can select a directory and electron saves the path onto the user config file under the attribute "downloadsDirectory".

In case we don't have permission to set some directory, electron throws an error and we use the default downloads directory.

In case the directory is deleted by the user, we create that during the app startup everytime.

screen shot 2017-11-16 at 09 02 07

### DIFF
--- a/js/menus/menuTemplate.js
+++ b/js/menus/menuTemplate.js
@@ -69,22 +69,24 @@ const template = [{
         click() {
             electron.dialog.showOpenDialog({
                 title: 'Select Downloads Directory',
-                defaultPath: '~/Downloads',
                 buttonLabel: 'Select',
                 properties: ['openDirectory', 'createDirectory']
             }, (filePaths) => {
+                if (!filePaths || !Array.isArray(filePaths) || filePaths.length < 1) {
+                    return;
+                }
                 updateConfigField('downloadsDirectory', filePaths[0]);
                 eventEmitter.emit('setDownloadsDirectory', filePaths[0]);
             });
         }
     },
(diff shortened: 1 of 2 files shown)


VishwasShashidhar:

js/menus/menuTemplate.js
```diff
@@ -65,7 +65,23 @@ const template = [{
...
     {
-        label: 'Open Crashes Directory',        
+        label: 'Set Downloads Directory',
+        click() {
+            electron.dialog.showOpenDialog({
+                title: 'Select Downloads Directory',
+                buttonLabel: 'Select',
+                properties: ['openDirectory', 'createDirectory']
+            }, (filePaths) => {
+                if (!filePaths || !Array.isArray(filePaths) || filePaths.length < 1) {
```
VikasShashidhar: So, does this happen when the user cancels the selection?

VishwasShashidhar: Yes, and when nothing is selected. It's just an extra check to ensure the sanctity of the app.

PR #239, opened 2017-11-16
Electron 214: Set "Minimize on Close" to true by default during installation

VishwasShashidhar:
[**Issue**](https://perzoinc.atlassian.net/browse/ELECTRON-214)

Currently, with the electron installer, the checkbox "Minimize on Close" is unchecked and will be false by default if we do not change the value during installation.

We need to reverse this and set the value of the checkbox to checked by default.

**Approach**

We change this behaviour in the Mac installer by checking the "Minimize on Close" checkbox in the Mac Installer pane and the Installer Dialog in the Advanced Installer.


VishwasShashidhar:

PR #245, opened 2017-11-23
SEARCH-480 (Message Filters)

keerthi16:
## Description
Converting the date from `string` to `int` to validate and added safety check
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SEARCH)

### DIFF
--- a/js/search/search.js
+++ b/js/search/search.js
@@ -256,16 +256,16 @@ class Search {
 
             let sd = new Date().getTime() - SEARCH_PERIOD_SUBTRACTOR;
             let sd_time = MINIMUM_DATE;
-            if (startDate && startDate !== "" && typeof startDate === 'object') {
-                sd_time = new Date(startDate).getTime();
+            if (startDate && startDate !== "") {
+                sd_time = new Date(parseInt(startDate, 10)).getTime();
                 if (sd_time >= sd) {
                     sd_time = sd;
                 }
             }
 
             let ed_time = MAXIMUM_DATE;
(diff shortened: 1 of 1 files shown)


js/search/search.js
```diff
@@ -254,18 +254,21 @@ class Search {
...
             }
 
-            let sd = new Date().getTime() - SEARCH_PERIOD_SUBTRACTOR;
-            let sd_time = MINIMUM_DATE;
-            if (startDate && startDate !== "" && typeof startDate === 'object') {
-                sd_time = new Date(startDate).getTime();
-                if (sd_time >= sd) {
-                    sd_time = sd;
+            let searchPeriod = new Date().getTime() - SEARCH_PERIOD_SUBTRACTOR;
+            let sd_time = searchPeriod;
```
VikasShashidhar: Please change the naming convention here. Underscore is ugly and the variable itself can be named to startDatePeriod. It's ok to have verbose names wherever necessary.

js/search/search.js
```diff
@@ -254,18 +254,21 @@ class Search {
...
                 }
             }
 
             let ed_time = MAXIMUM_DATE;
-            if (endDate && endDate !== "" && typeof endDate === 'object') {
-                ed_time = new Date(endDate).getTime();
+            if (endDate) {
+                let etime = new Date(parseInt(endDate, 10)).getTime();
+                if (etime) {
+                    ed_time = etime;
```
VikasShashidhar: Same as above.

PR #247, opened 2017-11-27
electron-219: added a pre-install script for macOS

VishwasShashidhar:
## Description
Currently, on macOS, when a new version of electron is installed, we are facing javascript issues, so, we have a ticket [ELECTRON-219](https://perzoinc.atlassian.net/browse/ELECTRON-219) which would address this by removing (quitting the app first) an already installed app on macOS during the mac installation process.

## Approach
- #### Problem with the code: Currently, there is no pre-install script
- #### Fix : Added a pre-install script which is part of the installer to quit & delete the already installed Symphony app.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

N/A

### DIFF
--- a/installer/mac/preinstall.sh
+++ b/installer/mac/preinstall.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+sudo killall Symphony
+sudo rm -rf /Applications/Symphony.app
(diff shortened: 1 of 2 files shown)


VishwasShashidhar:

PR #250, opened 2017-11-28
SEARCH-501 Search query should contain all tuples of the query tokens.

serkanmulayim:
## Description
In order to have better relevancy sorting search query should contain all tuples of the query.
For example if the query is "CAT? DOG, ELEPHANT! MONKEY$", the search query should be as following:
text"cat dog elephant monkey" "cat dog elephant" "dog elephant monkey" "cat dog" "dog elephant" "elephant monkey" "cat" "dog" "elephant" "monkey")

### DIFF
--- a/js/search/search.js
+++ b/js/search/search.js
@@ -370,8 +370,10 @@ class Search {
     static constructQuery(searchQuery, senderId, threadId, fileType) {
 
         let searchText = "";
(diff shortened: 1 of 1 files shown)


serkanmulayim:
ok to test

serkanmulayim:
ok to test

PR #251, opened 2017-11-28
Search 501

serkanmulayim:
## Description
In order to increase the relevancy we should provide the tuples of the entered terms separated by whitespaces. If a quotation mark exists in the query it implies a phrase search, then we should not change the query. Other punctuation marks are not going to be removed from the query and they will be stripped out by the tokenizer in the C library.

### DIFF
--- a/js/search/search.js
+++ b/js/search/search.js
@@ -474,11 +474,22 @@ class Search {
         return hashTags;
     }
 
+    /**
+     * If the search query does not have double quotes (implying phrase search),
+     * then create all tuples of the terms in the search query
+     * @param {String} searchText
+     * @returns {String}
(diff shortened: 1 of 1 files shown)


serkanmulayim:
1- if the query is CAT DOG ELEPHANT MONKEY, since there is no quotation marks, the search query should be as following:
text: ("cat dog elephant monkey" "cat dog elephant" "dog elephant monkey" "cat dog" "dog elephant" "elephant monkey" "cat" "dog" "elephant" "monkey")
2- If the query is "CAT DOG ELEPHANT MONKEY", then the query will be as following:
text: ("cat dog elephant monkey")

PR #273, opened 2017-12-27
electron-255: fixes electron-255

VishwasShashidhar:
## Description
As a result of electron-205, on Windows, when an item was asked to be "Show in Folder", the item was not being highlighted.

This was because we need to use backslashes and escape them on Windows. It has been fixed now.

## Approach
As per this ticket, found the fix -> https://github.com/electron/electron/issues/11362


VishwasShashidhar:

PR #274, opened 2017-12-28
electron-231: fixes the windows installer config setup issues.

VishwasShashidhar:
## Description
On Windows, during installation, if we uncheck "minimize on close" / "auto launch", the installer was not honoring the selection post install in the config file. Due to this, an user's selection was essentially not useful. The same has been fixed  [Electron-231](https://perzoinc.atlassian.net/browse/electron-231)

## Approach
- #### Problem with the code: In the "SetCheckboxValues" script, we weren't using the updated session variables.
- #### Fix: Used the updated session variable values.


VishwasShashidhar:

VishwasShashidhar:
I'll add this change as part of electron-205. Closing this PR now.

PR #280, opened 2017-12-29
Electron-222 (Fixes Notifications Javascript Error)

KiranNiranjan:
## Description
1. Fixes an error "Object has been destroyed"
2. Added method to close all notifications when the configuration is changed  

## Related Tickets
[JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-222)
[JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-254)

## Approach
- #### Problem with the code: There was no safety check added before accessing the browser window
- #### Fix: Added a safety check

### DIFF
--- a/js/notify/electron-notify.js
+++ b/js/notify/electron-notify.js
@@ -165,6 +165,7 @@ function updateConfig(customConfig) {
     if (customConfig.display) {
         displayId = customConfig.display;
     }
+    closeAll();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


js/notify/electron-notify.js
```diff
@@ -711,6 +719,34 @@ function cleanUpInactiveWindow() {
...
+function closeAll() {
+    // Clear out animation Queue and close windows
+    animationQueue.clear();
+
+    activeNotifications.forEach(function(window) {
+        if (window.displayTimer) {
+            clearTimeout(window.displayTimer);
+        }
+        if (window.electronNotifyOnCloseFunc) {
+            // ToDo: fix this: shouldn't delete method on arg
```
VikasShashidhar: Should I wait for this to be fixed?

KiranNiranjan: Done. Removed the block of code as it is not required.

PR #281, opened 2017-12-29
SEARCH-538 (Unit tests / Code refactoring)

keerthi16:
## Description
Unit tests for Electron Search and code refactoring. 


js/search/search.js
```diff
@@ -171,7 +171,7 @@ class Search {
     realTimeIndexing(message) {
         if (!message) {
             log.send(logLevels.ERROR, 'RealTime Indexing: Messages not provided');
-            return new Error('RealTime Indexing: Messages is required');
+            throw new Error('RealTime Indexing: Messages is required');
```
VikasShashidhar: I am curious here. Why do we make this check here? Why not in the calling method instead? Also, typo. Should be ```Messages are required```.

js/search/search.js
```diff
@@ -268,7 +268,7 @@ class Search {
 
             if (!fs.existsSync(this.indexFolderName) || !fs.existsSync(this.realTimeIndex)) {
                 log.send(logLevels.ERROR, 'Index folder does not exist.');
-                reject('Index folder does not exist.');
+                reject(new Error('Index folder does not exist.'));
```
VikasShashidhar: When we are throwing these errors, are we in any case indicating that the process failed? I think it would be good to do that, if at all it is discernible to the user that something like a "real time indexing" is happening.

tests/SearchUtils.test.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+});
+
+function mockedGetPath(type) {
+    if (type === 'exe') {
+        return executionPath;
+    }
+
+    if (type === 'userData') {
```
VikasShashidhar: Can use switch instead

tests/SearchUtils.test.js
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    afterAll(function (done) {
+        fs.unlinkSync(searchConfig.FOLDERS_CONSTANTS.USER_CONFIG_FILE);
+        done();
+    });
+
+    describe('Tests for checking disk space', function () {
+
+        it('should return free sapce', function (done) {
```
VikasShashidhar: Typo. ```space```

tests/SearchUtils.test.js
```diff
@@ -0,0 +1,149 @@
...
+                version: 1,
+                language: 'en'
+            };
+            SearchUtilsAPI.updateUserConfig(1234567891011, data).then(function (res) {
+                expect(res).toEqual(data);
+                done();
+            })
+        });
+
+        it('should modify user config file', function (done) {
```
VikasShashidhar: So, what is the difference between the above tests?

VikasShashidhar:
@keerthi16 We should expand our coverage for these tests. I'll merge this one once you make those corrections. We'll track that in a separate ticket later.

js/search/search.js
keerthi16: Just a safety check.

js/search/search.js
keerthi16: @VikasShashidhar We have a ticket for displaying indexing process and errors to the users. Will be handling this there 

Ticket Number : [SEARCH-551](https://perzoinc.atlassian.net/browse/SEARCH-551)

tests/SearchUtils.test.js
keerthi16: The first test is to update the user data which will be empty (the rotation-id will be 0).
In the second test changing the rotation-id and updating to config (after the updated rotation-id will be 1).

js/search/search.js
VikasShashidhar: 👍

js/search/search.js
VikasShashidhar: Yes, I understand. But why here instead of in the calling method? We could avoid a method call if it fails this check there.

tests/SearchUtils.test.js
VikasShashidhar: 👍

VikasShashidhar:
LGTM :+1

PR #282, opened 2017-12-29
Electron-231 (Fixes the installer config issue on Windows)

KiranNiranjan:
## Description
Fixed the regex in .aip to write the correct value to the user config  [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-231)

## Approach
- #### Problem with the code: Regex was looking for the `"minimizeOnClose";\s*:\s*false` instead of the value true
- #### Fix: Changed the regex to `"minimizeOnClose";\s*:\s*true`

### DIFF
--- a/installer/win/Symphony-x64.aip
+++ b/installer/win/Symphony-x64.aip
@@ -24,7 +24,7 @@
     <ROW Property="CTRLS" Value="2"/>
     <ROW Property="DialogBitmap" Value="dialog" MultiBuildValue="DefaultBuild:Tabloid.jpg" Type="1" MsiKey="DialogBitmap"/>
     <ROW Property="INVALID_POD_URL" Value="valid" Type="4"/>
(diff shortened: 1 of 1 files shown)


VishwasShashidhar:
lgtm 👍

So, basically because the default value was changed, we needed to update the regex right?

KiranNiranjan:
@VishwasShashidhar Yep :)

VikasShashidhar:
The JIRA ticket says that this issue happens for "Always on Top" as well. Should we not address that too?

VishwasShashidhar:
That's true. @KiranNiranjan Please do the needful.

KiranNiranjan:
@VikasShashidhar & @VishwasShashidhar The issue was only with the ```Minimize on Close```.

```Always on Top``` and ```Launch on Startup``` are working as expected.

PR #283, opened 2017-12-30
Electron-156 (Updated Protocol Validation)

KiranNiranjan:
## Description
Updated the code not to encode the URL and to just validate the protocol before creating the browser window
 [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-156)

## Approach
- #### Problem with the code: URL was encoding context-specific delimiters 
- #### Fix: Removed the encoding part of the code


VishwasShashidhar:
lgtm 👍

PR #284, opened 2018-01-02
electron-258: fixes the pop out issue for meetings and supports undefined urls in case of new window events as per the html dom specs

VishwasShashidhar:
## Description
Currently, in the electron app, if we try to open a new window with an undefined / blank url passed (which is supported as per the html dom specs), we do not open a new window and instead try to open it as an external link per the below code. [JIRA-ticket](https://perzoinc.atlassian.net/browse/electron-258)

`if (disposition === 'new-window' && newWinHost === mainWinHost) {`

So, due to this a lot of meetings related behaviour is affected in the electron app.

More details & discussions are in the JIRA ticket above.

## Approach
- #### Problem with the code: For urls with "about:blank", we try to open it in an external app which fails on the OS side and we end up with not opening the link in either electron or an external app.
- #### Fix: We whitelist "about:blank" as part of the above mentioned if condition like below:

`if (disposition === 'new-window' && ((newWinHost === mainWinHost) || newWinUrl === 'about:blank'))`

## Learning
html dom specs allows us to open a window without any url (i.e. "about:blank:) https://www.w3schools.com/jsref/met_win_open.asp


VishwasShashidhar:

CC: @lneir @KiranNiranjan

js/windowMgr.js
```diff
@@ -320,7 +320,7 @@ function doCreateMainWindow(initialUrl, initialBounds) {
 
         // only allow window.open to succeed is if coming from same hsot,
         // otherwise open in default browser.
-        if (disposition === 'new-window' && newWinHost === mainWinHost) {
+        if (disposition === 'new-window' && ((newWinHost === mainWinHost) || newWinUrl === 'about:blank')) {
```
VikasShashidhar: I would rather this is stored in a variable and that checked here instead of hardcoding it.

VishwasShashidhar: Done

PR #290, opened 2018-01-10
Electron-262 (Fixes Badge Count Issue)

KiranNiranjan:
## Description
Implemented a method to reset badge count on window navigation/reload [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-262)

## Approach
- #### Problem with the code: Badge count was not being reset to 0 when reloaded/navigated
- #### Fix: Added an event to reset badge count whenever the window is reloaded/navigated

@VikasShashidhar & @VishwasShashidhar Please review. Thanks 😄

### DIFF
--- a/js/preload/preloadMain.js
+++ b/js/preload/preloadMain.js
@@ -367,8 +367,17 @@ function createAPI() {
         });
     }
 
+    // reset the badge count when ever user refresh
+    function resetBadgeCount() {
+        local.ipcRenderer.send(apiName, {
+            cmd: apiCmds.setBadgeCount,
+            count: 0
+        });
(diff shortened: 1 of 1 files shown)


js/preload/preloadMain.js
```diff
@@ -367,8 +367,17 @@ function createAPI() {
         });
     }
 
+    // reset the badge count when ever user refresh
```
VishwasShashidhar: Please change the comment to below
`// reset the badge count whenever an user refreshes the electron client`

VishwasShashidhar:
lgtm 👍

PR #292, opened 2018-01-18
Electron-228 (Fixes Screen Snippet Reload Issue)

KiranNiranjan:
## Description
Added an event to terminate screen snippet process whenever the main window is reloaded/navigated [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-228)

Also updated the badge count feature so that there will be one reload event to handle everything.

## Approach
- #### Problem with the code: Screen snippet was not being terminated when the app is reloaded as the context will be lost
- #### Fix: Added an event to terminate screen snippet whenever the main window is reloaded/navigated

@VikasShashidhar & @VishwasShashidhar Please review. Thanks 😄

### DIFF
--- a/js/mainApiMgr.js
+++ b/js/mainApiMgr.js
@@ -13,6 +13,8 @@ const activityDetection = require('./activityDetection');
 const badgeCount = require('./badgeCount.js');
 const protocolHandler = require('./protocolHandler');
 const configureNotification = require('./notify/settings/configure-notification-position');
+const eventEmitter = require('./eventEmitter');
+const { isMac } = require('./utils/misc');
 
 const apiEnums = require('./enums/api.js');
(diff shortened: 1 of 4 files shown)


js/mainApiMgr.js
```diff
@@ -47,6 +49,24 @@ function isValidWindow(event) {
     return result;
 }
 
+/**
+ * Method that is invoked when the application is reloaded/navigated
+ * window.addEventListener('beforeunload')
+ * @param windowName
+ */
+function reload(windowName) {
```
VikasShashidhar: Sorry I am bringing this up late, but we can use a name such as ```resetNecessities``` or ```sanitize``` or something along those lines. ```reload``` to me sounds like something else is happening.

PR #294, opened 2018-01-24
electron-261: add gpu switches

VishwasShashidhar:
## Description
For windows environments to ensure screen-sharing doesn't break with multiple monitors, we add flags to disable gpu
 [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-261)

## Approach
With electron, we can set chrome flags which should help us with the issue.

## Learning
https://windowsreport.com/chrome-black-screen-problem-windows-10/

### DIFF
--- a/config/Symphony.config
+++ b/config/Symphony.config
@@ -17,8 +17,6 @@
     "customFlags": {
         "authServerWhitelist": "",
         "authNegotiateDelegateWhitelist": "",
-        "disableGpu": false,
-        "disableGpuCompositing": false,
-        "disableD3d11": false
+        "disableGpu": false
     }
 }
--- a/config/Symphony.config
+++ b/config/Symphony.config
@@ -16,6 +16,9 @@
     },
     "customFlags": {
         "authServerWhitelist": "",
-        "authNegotiateDelegateWhitelist": ""
+        "authNegotiateDelegateWhitelist": "",
+        "disableGpu": false,
+        "disableGpuCompositing": false,
+        "disableD3d11": false
     }
 }
--- a/js/main.js
+++ b/js/main.js
@@ -13,7 +13,7 @@ const urlParser = require('url');
 // Local Dependencies
 const {getConfigField, updateUserConfigWin, updateUserConfigMac, readConfigFileSync} = require('./config.js');
 const {setCheckboxValues} = require('./menus/menuTemplate.js');
-const { isMac, isDevEnv } = require('./utils/misc.js');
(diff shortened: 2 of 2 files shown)


VishwasShashidhar:

CC: @lneir

VishwasShashidhar:
[electron-261 test run.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1659948/electron-261.test.run.pdf)

Adding unit tests results.

js/main.js
```diff
@@ -22,6 +22,15 @@ const { deleteIndexFolder } = require('./search/search.js');
 
 require('electron-dl')();
 
+// ELECTRON-261: On Windows, due to gpu issues, we need to disable gpu
+// to ensure screen sharing works effectively with multiple monitors
+// https://github.com/electron/electron/issues/4380
+if (isWindowsOS) {
+    app.commandLine.appendSwitch("disable-gpu", true);
```
lneir: so we are doing this always???  isn't the problem only related to screensharing?  is there a way to only disable when sharing a screen?  what is the perf impact to user that have gpu - will the electron wrapper perform more poorly. 

also does this happen in all versions of os - or just win10 - why not then just do for win10?

maybe another option is selectively allow users to disable?  maybe a user feature or an install option???

VishwasShashidhar: 1. This cannot be specifically turned on or off for a screen sharing session because the flags only get enabled before the "on-ready" event from the main process.

2. I'll profile the app and get back with the performance metrics.

3. Makes sense to limit this to Windows 10. Will make necessary changes.

4. Great idea to have this selectively enabled by the administrators from the config file. Will check with product to take their input.

Thanks for the feedback.

VikasShashidhar:
@lneir Can you please review this again and confirm if we are good to merge? Thanks.

config/Symphony.config
```diff
@@ -16,6 +16,9 @@
     },
     "customFlags": {
         "authServerWhitelist": "",
-        "authNegotiateDelegateWhitelist": ""
+        "authNegotiateDelegateWhitelist": "",
+        "disableGpu": false,
```
lneir: do we need such granular configuration?  
that is, are there cases where we just need to disable one but not others?

VishwasShashidhar: agreed. makes sense just to have one attribute. will push the change.

VikasShashidhar:
@lneir Are we good to go with this?

VishwasShashidhar:
Closing this to open a PR with the current branch i.e. ELECTRON-271

PR #298, opened 2018-02-06
SEARCH-605 (Electron - Cannot enable swift search on Windows 7)

keerthi16:
## Description
Electron - Cannot enable swift search on Windows 7
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SEARCH-605)

## Approach
- #### Problem with the code: Windows 7 had permission issue for getting the available disk space 
- #### Fix: Created a new custom execution file which will not throw any permission issue.

Build passing electron-custom-build # 13
screen shot 2018-02-06 at 11 53 22 am

- [X] Unit-Tests
- [X] Documentation
- [ ] Automation-Tests


js/search/utils/checkDiskSpace.js
```diff
@@ -25,17 +25,25 @@ function checkDiskSpace(path, resolve, reject) {
...
-                if (stderr.indexOf("No such file or directory") !== -1) {
+                if (stderr.indexOf("The system cannot find the path specified.") !== -1) {
                     return reject(new Error("No such file or directory : " + error));
                 }
                 return reject(new Error("Error : " + error));
             }
-            let data = stdout.trim().split("\n");
+            let data = stdout.trim().split(",");
+
+            if (data[ 1 ] === "NOTREADY") {
```
VikasShashidhar: Please move these to enums.

js/search/utils/checkDiskSpace.js
```diff
@@ -25,17 +25,25 @@ function checkDiskSpace(path, resolve, reject) {
...
+                return reject(new Error("Error : Disk not ready"));
+            }
+
+            if (data[ 1 ] === "DISKNOTFOUND") {
```
VikasShashidhar: Same with this.

installer/win/Symphony-x64.aip
```diff
@@ -126,6 +126,7 @@
     <ROW Component="cld.node" ComponentId="{D9F74D7B-B2E4-4ED8-A023-A54A98AE3F96}" Directory_="Release_Dir" Attributes="256" KeyPath="cld.node" Type="0"/>
     <ROW Component="d3dcompiler_47.dll" ComponentId="{C7B87C02-3116-43A8-A70B-3592B70E6AC8}" Directory_="APPDIR" Attributes="256" KeyPath="d3dcompiler_47.dll"/>
     <ROW Component="ffmpeg.dll" ComponentId="{A1C4A332-3490-44D8-A5C9-9523889B488B}" Directory_="APPDIR" Attributes="256" KeyPath="ffmpeg.dll"/>
+    <ROW Component="freediskspace.exe" ComponentId="{012071B8-758A-4A24-8A13-A48476F9F6FC}" Directory_="library_Dir" Attributes="0" KeyPath="freediskspace.exe"/>
```
VikasShashidhar: Is there no other way to do this? If not, what operations are this file doing? Can we not use a DLL instead?

@VishwasShashidhar @lneir Can you suggest an alternative if any?

keerthi16: @VikasShashidhar I was using Windows 'Fsutil' before but it had permission issues so thought of using the `DriveInfo.AvailableFreeSpace Property` and created an execution file and now this won't have any permission issue.

js/search/searchConfig.js
```diff
@@ -26,12 +26,15 @@ const libraryPath = isMac ? path.join(macLibraryPath, 'libsymphonysearch.dylib')
 const userConfigFileName = 'search_users_config.json';
 const userConfigFile = isDevEnv ? path.join(__dirname, '..', '..', userConfigFileName) : path.join(userData, userConfigFileName);
 
+const freeDiskSpace = path.join(winLibraryPath, 'free-disk-space.exe');
```
lneir: i see we are checking in a binary file.
why do we need it?  what does it do?
where are we getting this binary file from?
do we trust it?
assume only runs on windows???
will it run on all windows platforms?  does it have DLL dependencies?

keerthi16: @lneir 
1. Before I was using the `fsutils` Which requires admin rights to execute (details in this [link](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/fsutil))
2. Here is the [link](https://msdn.microsoft.com/en-us/library/system.io.driveinfo.availablefreespace(v=vs.110).aspx?cs-save-lang=1&cs-lang=csharp#code-snippet-2) for checking the `AvailableFreeSpace` code in C#.
I'm using this code and made some changes as per required and compiled using `csc` [link](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/command-line-building-with-csc-exe).
3. And the binary only check for free disk space for the provided drive name.
4. As it does not require any dependency and only using system.io which works on all windows platform.

lneir: where is the source code?  we should have this in a repo and have builds setup to provide exe.

PR #301, opened 2018-02-15
SEARCH-605 (Electron - Cannot enable swift search on Windows 7)

keerthi16:
## Description
Electron - Cannot enable swift search on Windows 7
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SEARCH-605)

_Note: This is a temporary patch. If there are permission issue and the disk space is less than 300MB the swift search will be enabled._

## Approach
- #### Problem with the code: FSUTIL utility requires administrative privileges
- #### Fix: This fix will be released as a part of [SEARCH-621](https://perzoinc.atlassian.net/browse/SEARCH-621)

## Learning
[Fsutil](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc753059(v=ws.10))

- [x] Unit-Tests
screen shot 2018-02-15 at 12 31 31 pm
- [x] Documentation
- [ ] Automation-Tests

### DIFF
--- a/js/search/utils/checkDiskSpace.js
+++ b/js/search/utils/checkDiskSpace.js
@@ -32,13 +32,13 @@ function checkDiskSpace(path, resolve, reject) {
                 }
                 if (stdout.indexOf("The FSUTIL utility requires that you have administrative privileges.") !== -1) {
                     // this is temporary until we use the custom exe file.
-                    return true;
+                    return resolve(true);
                 }
                 return reject(new Error("Error : " + error));
             }
             if (stdout.indexOf("The FSUTIL utility requires that you have administrative privileges.") !== -1) {
                 // this is temporary until we use the custom exe file.
-                return true;
+                return resolve(true);
             }
             let data = stdout.trim().split("\n");
 
--- a/js/search/utils/checkDiskSpace.js
+++ b/js/search/utils/checkDiskSpace.js
@@ -25,12 +25,20 @@ function checkDiskSpace(path, resolve, reject) {
             return resolve(space >= searchConfig.MINIMUM_DISK_SPACE);
         });
     } else {
-        exec(`fsutil volume diskfree ${path}`, (error, stdout, stderr) => {
+        exec(`fsutil volume diskfree ${path}`, (error, stdout) => {
             if (error) {
-                if (stderr.indexOf("No such file or directory") !== -1) {
+                if (stdout.indexOf("Error:  The system cannot find the path specified.") !== -1) {
                     return reject(new Error("No such file or directory : " + error));
                 }
+                if (stdout.indexOf("The FSUTIL utility requires that you have administrative privileges.") !== -1) {
+                    // this is temporary until we use the custom exe file.
+                    return true;
+                }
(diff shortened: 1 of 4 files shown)


js/search/utils/checkDiskSpace.js
```diff
@@ -25,13 +25,21 @@ function checkDiskSpace(path, resolve, reject) {
             return resolve(space >= searchConfig.MINIMUM_DISK_SPACE);
         });
     } else {
-        exec(`fsutil volume diskfree ${path}`, (error, stdout, stderr) => {
+        exec(`fsutil volume diskfree ${path}`, (error, stdout) => {
             if (error) {
-                if (stderr.indexOf("No such file or directory") !== -1) {
+                if (stdout.indexOf("Error:  The system cannot find the path specified.") !== -1) {
```
VikasShashidhar: Move this out to a const or enum please.

js/search/utils/checkDiskSpace.js
```diff
@@ -25,13 +25,21 @@ function checkDiskSpace(path, resolve, reject) {
...
                     return reject(new Error("No such file or directory : " + error));
                 }
+                if (stdout.indexOf("The FSUTIL utility requires that you have administrative privileges.") !== -1) {
```
VikasShashidhar: Same here.

js/search/utils/checkDiskSpace.js
VikasShashidhar: Can we not check this against a flag or something instead of the error message? It seems a little wrong to me.

js/search/utils/checkDiskSpace.js
```diff
@@ -25,13 +25,21 @@ function checkDiskSpace(path, resolve, reject) {
...
+                    // this is temporary until we use the custom exe file.
+                    return resolve(true);
+                }
                 return reject(new Error("Error : " + error));
             }
+            if (stdout.indexOf("The FSUTIL utility requires that you have administrative privileges.") !== -1) {
```
VikasShashidhar: Same here.

tests/SearchUtils.test.js
```diff
@@ -33,6 +33,9 @@ describe('Tests for Search Utils', function() {
 
     beforeAll(function (done) {
         executionPath = path.join(__dirname, 'library');
+        if (!isMac) {
```
VikasShashidhar: This is probably a wrong check if we want it to work on Linux in the future. Is the additional `..` required only for Windows only?

keerthi16: @VikasShashidhar 
Mac path `executionPath = path.join(__dirname, 'library');`
Windows path `executionPath = path.join(__dirname, '..', 'library');`

In Windows, the library is present in the same dir as the `.exe` path.
But in Mac library is present outside of the dir.

This is the config file for which I'm passing the paths.
screen shot 2018-02-16 at 1 25 35 pm

VikasShashidhar:
@keerthi16 Please resolve conflicts

keerthi16:
@VikasShashidhar conflicts resolved.

PR #304, opened 2018-02-19
Electron 276: add gpu flags in electron

VishwasShashidhar:
## Description
For windows environments to ensure screen-sharing doesn't break with multiple monitors, we add flags to disable gpu
 [JIRA-ticket](https://perzoinc.atlassian.net/browse/ELECTRON-276)

## Approach
With electron, we can set chrome flags which should help us with the issue.

## Learning
https://windowsreport.com/chrome-black-screen-problem-windows-10/

### DIFF
--- a/config/Symphony.config
+++ b/config/Symphony.config
@@ -17,8 +17,6 @@
     "customFlags": {
         "authServerWhitelist": "",
         "authNegotiateDelegateWhitelist": "",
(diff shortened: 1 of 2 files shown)


VishwasShashidhar:
@lneir This should be the correct branch / PR for the gpu flags requirement.

PR #308, opened 2018-02-21
Electron-256 (Changed the implementation from modal to electron window)

KiranNiranjan:
## Description
Previous implementation (`MediaSelectorModal`) had issues with the screen picker not showing when the meeting is in fullscreen mode. Changed the implementation from modal to electron browser window 

#### This PR fixes the following Tickets:
[ELECTRON-256](https://perzoinc.atlassian.net/browse/ELECTRON-256)
[ELECTRON-197](https://perzoinc.atlassian.net/browse/ELECTRON-197)

## Approach
- #### Problem with the code: Clicking on the `share screen` button in the meeting does not show the screen picker modal when the meeting is in fullscreen mode.
- #### Fix:  Changing the implementation from modal to electron browser window fixes the issue

#### Before
screen shot 2018-02-02 at 10 08 32 am

#### After
screen shot 2018-02-21 at 2 39 21 pm

## Learning
N/A

#### Blog Posts
N/A

- [ ] Unit-Tests (Not Applicable)
- [ ] Documentation (No Changes)
- [ ] Automation-Tests

@VikasShashidhar & @VishwasShashidhar Please Review. Thanks 😄

### DIFF
--- a/js/desktopCapturer/getSource.js
+++ b/js/desktopCapturer/getSource.js
@@ -13,7 +13,11 @@
 // electron: https://github.com/electron/electron/issues/9312
 
 const { ipcRenderer, remote } = require('electron');
+const apiEnums = require('../enums/api.js');
+const apiCmds = apiEnums.cmds;
+const apiName = apiEnums.apiName;
 const { isWindowsOS } = require('../utils/misc');
+const USER_CANCELLED = 'User Cancelled';
 
 let nextId = 0;
 let includes = [].includes;
@@ -37,7 +41,7 @@ function isValid(options) {
  * @param callback
  * @returns {*}
  */
-function getSources(options, callback) {
+function getSource(options, callback) {
     let captureScreen, captureWindow, id;
     if (!isValid(options)) {
         return callback(new Error('Invalid options'));
@@ -68,27 +72,25 @@ function getSources(options, callback) {
     ipcRenderer.send('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_SOURCES', captureWindow, captureScreen, updatedOptions.thumbnailSize, id);
 
     return ipcRenderer.once('ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_' + id, function(event, sources) {
-        let source;
-        callback(null, (function() {
-            let i, len, results;
-            results = [];
-            for (i = 0, len = sources.length; i < len; i++) {
-                source = sources[i];
-                results.push({
-                    id: source.id,
-                    name: source.name,
-                    thumbnail: source.thumbnail
-                });
+
+        ipcRenderer.send(apiName, {
+            cmd: apiCmds.openScreenPickerWindow,
+            sources: sources,
+            id: id
+        });
+
+        function successCallback(e, source) {
+            // Cleaning up the event listener to prevent memory leaks
+            if (!source) {
+                ipcRenderer.removeListener('start-share' + id, func);
+                return callback(new Error(USER_CANCELLED));
             }
+            return callback(null, source);
+        }
 
-            return results;
-
-        }()));
+        const func = successCallback.bind(this);
+        ipcRenderer.once('start-share' + id, func);
     });
 }
 
-/**
- * @deprecated instead use getSource
- * @type {getSources}
- */
-module.exports = getSources;
+module.exports = getSource;
--- a/js/desktopCapturer/getSources.js
+++ b/js/desktopCapturer/getSources.js
@@ -13,6 +13,9 @@
 // electron: https://github.com/electron/electron/issues/9312
 
 const { ipcRenderer, remote } = require('electron');
+const apiEnums = require('../enums/api.js');
+const apiCmds = apiEnums.cmds;
+const apiName = apiEnums.apiName;
 const { isWindowsOS } = require('../utils/misc');
 
 let nextId = 0;
@@ -67,9 +70,15 @@ function getSources(options, callback) {
     id = getNextId();
     ipcRenderer.send('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_SOURCES', captureWindow, captureScreen, updatedOptions.thumbnailSize, id);
 
-    return ipcRenderer.once('ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_' + id, function(event, sources) {
-        let source;
-        callback(null, (function() {
+    ipcRenderer.once('ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_' + id, function(event, sources) {
+        //let source;
+
+        ipcRenderer.send(apiName, {
+            cmd: apiCmds.openScreenPickerWindow,
+            sources: sources
+        });
+
+        /*callback(null, (function() {
             let i, len, results;
             results = [];
             for (i = 0, len = sources.length; i < len; i++) {
@@ -83,8 +92,14 @@ function getSources(options, callback) {
 
             return results;
 
-        }()));
+        }()));*/
+
+        return ipcRenderer.on('screen-selected', function (e, source) {
+            console.error("here also");
+            callback(null, source);
+        })
     });
+
 }
 
 module.exports = getSources;
--- a/js/desktopCapturer/getSources.js
+++ b/js/desktopCapturer/getSources.js
@@ -71,7 +71,7 @@ function getSources(options, callback) {
     id = getNextId();
     ipcRenderer.send('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_SOURCES', captureWindow, captureScreen, updatedOptions.thumbnailSize, id);
 
-    ipcRenderer.once('ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_' + id, function(event, sources) {
+    return ipcRenderer.once('ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_' + id, function(event, sources) {
 
         ipcRenderer.send(apiName, {
             cmd: apiCmds.openScreenPickerWindow,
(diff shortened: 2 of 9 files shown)


VishwasShashidhar:
I think we should remove link to ELECTRON-310 because the fix was applied to R50 anyway.

js/desktopCapturer/getSources.js
```diff
@@ -68,22 +72,24 @@ function getSources(options, callback) {
...
+        ipcRenderer.send(apiName, {
+            cmd: apiCmds.openScreenPickerWindow,
+            sources: sources,
+            id: id
+        });
+
+        function successCallback(e, source) {
+            // Cleaning up the event listener to prevent memory leaks
+            if (!source) {
+                ipcRenderer.removeListener('stat-share' + id, func);
```
VishwasShashidhar: `stat-share`, is that a typo?

js/desktopCapturer/index.js
```diff
@@ -0,0 +1,140 @@
...
+const log = require('../log.js');
+const logLevels = require('../enums/logLevels.js');
+const { isMac, isWindowsOS } = require('./../utils/misc.js');
+
+let screenPickerWindow;
+let preloadWindow;
+let eventId;
+
+let windowConfig = {
+    width: 580,
```
VishwasShashidhar: can we get this dynamically?

js/desktopCapturer/index.js
```diff
@@ -0,0 +1,140 @@
...
+    height: isMac ? 519 : 523,
```
VishwasShashidhar: can we get these dynamically?

js/desktopCapturer/index.js
```diff
@@ -0,0 +1,140 @@
...
+    show: false,
+    modal: true,
+    frame: false,
+    autoHideMenuBar: true,
+    resizable: false,
+    alwaysOnTop: true,
+    webPreferences: {
+        preload: path.join(__dirname, 'renderer.js'),
+        sandbox: true,
```
VishwasShashidhar: is this needed?

js/desktopCapturer/index.js
```diff
@@ -0,0 +1,140 @@
...
+    screenPickerWindow = new BrowserWindow(windowConfig);
+    screenPickerWindow.setVisibleOnAllWorkspaces(true);
+    screenPickerWindow.loadURL(getTemplatePath());
+
+    screenPickerWindow.once('ready-to-show', () => {
+        screenPickerWindow.show();
+    });
+
+    screenPickerWindow.webContents.on('did-finish-load', () => {
+        screenPickerWindow.webContents.send('desktop-capturer-sources', sources, isWindowsOS);
```
VishwasShashidhar: what is the reason for checking windows os?

js/desktopCapturer/renderer.js
```diff
@@ -0,0 +1,264 @@
+'use strict';
+const { ipcRenderer } = require('electron');
+
+const screenRegExp = new RegExp(/^Screen \d+$/gmi);
+const PAGE_DOWN = 34,
```
VishwasShashidhar: can we add code comments for these?

js/desktopCapturer/getSources.js
KiranNiranjan: Yes. I've fixed it 👍

js/desktopCapturer/index.js
KiranNiranjan: Dynamically with reference to display resolution or HTML content?

js/desktopCapturer/index.js
KiranNiranjan: `isWindowsOS`is required to add the window borders only when the app is running on Windows OS because creating a frameless window electron app loses its native borders and drop shadow.

js/desktopCapturer/index.js
KiranNiranjan: Yes, It makes this `browserWindow` more secure 😄

js/desktopCapturer/renderer.js
```diff
@@ -0,0 +1,265 @@
...
+    shareButton.innerText = 'Share'
+}
+
+/**
+ * Loops through the items and removes
+ * the selected class property
+ */
+function highlightSelectedSource() {
+    let items = document.getElementsByClassName('item-container');
+    for (let i = 0, len = items.length; i < len; i++) {
```
VikasShashidhar: Could be simplified to `for (let i = 0; i < items.length; i++)`

js/desktopCapturer/renderer.js
```diff
@@ -0,0 +1,265 @@
...
+    if (shareButton && shareButton.classList[0] === 'share-button-disable') {
+        shareButton.innerText = text;
+    }
+}
+
+/**
+ * Method handles used key up event
+ * @param e
+ */
+function handleKeyUpPress(e) {
```
VikasShashidhar: Could be renamed to `handleKeyboardInteractions`

js/desktopCapturer/renderer.js
```diff
@@ -0,0 +1,265 @@
...
+                updateSelectedSource(availableSources.length - 1);
+            }
+            break;
+        case ARROW_DOWN:
+            updateSelectedSource(2);
+            break;
+        case ENTER_KEY:
+            startShare();
+            break;
+        case ESCAPE_KEY:
```
VikasShashidhar: We could also consider implementing for `tab`

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Screen Picker</title>
+    <style>
+
+        body {
+            margin: 0 auto;
+            font-family: "Helvetica Neue", sans-serif;
```
VikasShashidhar: Can we look at following the font fallback used in the client app?

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
...
+            overflow: hidden;
+        }
+
+        .window-border {
+            border: 2px solid #444;
```
VikasShashidhar: Try to use the RGBA color coding. You'll get finer control

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
...
+
+        span {
+            font-style: normal;
+            margin: 0 0 0 4px;
+            font-size: 1.2rem;
+            min-height: 13px;
+        }
+
+        input {
+            display: none;
```
VikasShashidhar: Why is this required?

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
...
+            display: inline-block;
+            height: 150px;
+            margin: 5px;
+            position: relative;
+            width: 40%;
+            border: 1px solid rgba(0, 0, 0, .1);
+        }
+
+        .item-container:hover {
+            box-shadow: 0 1px 5px rgba(0, 0, 0, .1), 0 5px 5px rgba(0, 0, 0, .1);
```
VikasShashidhar: You could also make the cursor a pointer.

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
...
+            color: rgba(0, 0, 0, 0.38);
+            background-color: rgba(255, 255, 255, 0);
+            text-transform: uppercase;
+            margin: 0;
+            padding: 8px 16px;
+        }
+
+        .cancel-button:hover {
+            background-color: rgba(255, 255, 255, 0);
+            color: black;
```
VikasShashidhar: Could use a cursor pointer. Better indicates to the user that it is actionable.

js/desktopCapturer/screen-picker.html
```diff
@@ -0,0 +1,239 @@
...
+        #error-content span {
+            color: rgba(0, 0, 0, 0.38);
+        }
+
+    </style>
+</head>
+<body>
+<div class="content">
+    <div class="custom-window-title">
+        <span>Choose what you'd like to share</span>
```
VikasShashidhar: Try to use HTML unicodes for symbols - https://www.fileformat.info/info/unicode/char/27/index.htm

js/desktopCapturer/index.js
KiranNiranjan: For creating a dynamic window with respect to HTML content. first, we need to open the window with some value for `width & height` and then we need to publish an event from `render process` to the `main process` when the DOM is loaded to resize the window.

Currently, the `width and height` of the window is exactly the same as the HTML content. I don't think that we need to dynamically update the window `width or height`

js/desktopCapturer/renderer.js
KiranNiranjan: This method will be invoked only for `keyup` events

js/desktopCapturer/screen-picker.html
KiranNiranjan: Input type for the tabs that I'm using is `radio` so in order to hide the radio button setting `display: none`

KiranNiranjan:
@VikasShashidhar & @VishwasShashidhar Added a new API to support backward compatibility and deprecated the existing one. Please review. Thanks 😄

js/desktopCapturer/getSource.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
```
VikasShashidhar: Is this just a new API that does the same thing as the `getSources` API? If yes, why reinvent it?

KiranNiranjan: @VikasShashidhar `getSources` returns an array of sources whereas `getSource` returns single source selected by the user. We are keeping the `getSources` API to support older versions of Electron and Client app.

We can remove `getSources` after 2-3 sprints 😄

PR #341, opened 2018-04-19
Electron-433 (Changes the logic for updating user config)

KiranNiranjan:
## Description
Changed the logic from updating the user config from launching an instance of the Symphony to when the app is first launched [ELECTRON-433](https://perzoinc.atlassian.net/browse/ELECTRON-433)

## Approach
#### Problem with the code: 
 - The previous implementation had a permissions issue with the app path being `root`
#### Fix:
 - Changed the logic to update the user config on the first launch

- [X] Unit-Tests
- [X] Documentation
- [X] Automation-Tests

@VishwasShashidhar @VikasShashidhar Please review. Thanks 😄

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -21,7 +21,7 @@ const configFileName = 'Symphony.config';
 let userConfig;
 let globalConfig;
 
-let ignoreSettings = ['minimizeOnClose', 'launchOnStartup', 'alwaysOnTop', 'url', 'memoryRefresh', 'bringToFront', 'firstTimeLaunch'];
+let ignoreSettings = ['minimizeOnClose', 'launchOnStartup', 'alwaysOnTop', 'url', 'memoryRefresh', 'bringToFront'];
 
 /**
  * Tries to read given field from user config file, if field doesn't exist
--- a/js/config.js
+++ b/js/config.js
@@ -4,7 +4,6 @@ const electron = require('electron');
 const app = electron.app;
 const path = require('path');
 const fs = require('fs');
-const AppDirectory = require('appdirectory');
 const omit = require('lodash.omit');
 const pick = require('lodash.pick');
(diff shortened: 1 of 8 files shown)


js/config.js
```diff
@@ -16,14 +15,13 @@ const log = require('./log.js');
...
 const configFileName = 'Symphony.config';
-const dirs = new AppDirectory('Symphony');
 
 // cached config when first reading files. initially undefined and will be
 // updated when read from disk.
 let userConfig;
 let globalConfig;
 
-let ignoreSettings = ['minimizeOnClose', 'launchOnStartup', 'alwaysOnTop', 'url', 'memoryRefresh', 'bringToFront'];
+let ignoreSettings = ['minimizeOnClose', 'launchOnStartup', 'alwaysOnTop', 'url', 'memoryRefresh', 'bringToFront', 'firstTimeLaunch'];
```
VishwasShashidhar: is `firstTimeLaunch` required here?

js/main.js
```diff
@@ -219,45 +226,67 @@ function setupThenOpenMainWindow() {
...
+                const shouldUpdateUserConfig = execPath.indexOf('AppData/Local/Programs') !== -1 || isMac;
+
+                if (configVersion
+                    && typeof configVersion === 'string'
+                    && (compareSemVersions.check(appVersionString, configVersion) === 0
+                    || compareSemVersions.check(appVersionString, configVersion) === -1)) {
+                    return resolve();
+                }
+
+                if (shouldUpdateUserConfig) {
```
VishwasShashidhar: can we combine the comparison of version check logic with the exe path and have one single if condition?

js/config.js
KiranNiranjan: No, Removed 👍

PR #347, opened 2018-04-30
Electron 385: add support for custom tlds

VishwasShashidhar:
## Description
Currently, in case of custom tlds, the "unauthenticated" dialog pops up as per the ticket [ELECTRON-385](https://perzoinc.atlassian.net/browse/ELECTRON-385), this fix adds a new parameter in the Symphony.config file to be able to add custom tlds which would validate a url in the custom TLD.

## Approach
- #### Problem with the code: The "whitelistHandler" code doesn't currently support validating custom TLDs
- #### Fix: Add a new parameter to the "Symphony.config" file where an admin can add custom TLDs to be whitelisted. Read this in the whitelistHandler and pass it to the `parseDomain` function which would consider the custom TLD while validating the URL.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

- [x] Unit-Tests
[ELECTRON-385 - Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1960401/ELECTRON-385.-.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-385 - Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1960403/ELECTRON-385.-.Spectron.Tests.pdf)


js/utils/whitelistHandler.js
```diff
@@ -35,7 +52,7 @@ function isWhitelisted(url) {
  */
 function checkWhitelist(url, whitelist) {
     let whitelistArray = whitelist.split(',');
-    const parsedUrl = parseDomain(url);
+    const parsedUrl = parseDomain(url, {customTlds});
```
lneir: seems like having to add this customTlds is not a great way to go - just yet another config item for customer to have to think about and possibly cause confusion.  and also seems like it is just because this lib 'parseDomain' being used is not that great.  seems like writing our code might be better way to go, likely even simple regex could be used.

VishwasShashidhar: makes sense, will remove the dependency and write our own code. but, validating the tlds will have to be removed because we'll have to maintain a list of them to validate against which would come back to the same point. in any case, since we are validating against the whitelist array, should be fine even in case of a harmful url.

VishwasShashidhar: addressed in https://github.com/symphonyoss/SymphonyElectron/pull/350

PR #358, opened 2018-05-07
Electron 466: fix setting user data path using command line

VishwasShashidhar:
## Description
To run automation tests, the automation suite sets custom user data path for each electron instance in case of launching multiple instances. This needs to be re-enabled to fix regression. [ELECTRON-466](https://perzoinc.atlassian.net/browse/ELECTRON-466)

## Approach
- #### Problem with the code: Fixing ELECTRON-433 caused a regression where the user data path code was removed.
- #### Fix: Added the code back for parsing command line parameter to set user data path.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

- [x] Unit-Tests
[ELECTRON-466 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1979344/ELECTRON-466.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-466 Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1979343/ELECTRON-466.Spectron.Tests.pdf)


js/main.js
```diff
@@ -228,6 +228,16 @@ function setupThenOpenMainWindow() {
     isAppAlreadyOpen = true;
     getUrlAndCreateMainWindow();
     
+    // Allows a developer to set custom user data path from command line when
+    // launching the app. Mostly used for running automation tests with
+    // multiple instances
+    let customDataArg = getCmdLineArg(process.argv, '--userDataPath=', false);
+    
+    if (customDataArg && customDataArg.split('=').length > 1) {
+        let customDataFolder = customDataArg.split('=')[1];
```
KiranNiranjan: @VishwasShashidhar I don't think this would work if the path has '='

ex: --userDataPath="C:\Users\Test_user\Desktop\n=ote.txt"
will result in [ '--userDataPath', '"C:\\Users\Test_user\Desktop\n', 'ote.txt"' ]

VishwasShashidhar: as confirmed by Daniel, they just pass the "path" of the folder and not the filename itself.

screen shot 2018-05-07 at 15 43 29

VikasShashidhar: I agree with @KiranNiranjan here. Maybe we could identify the first "=" with a regex and then get a substring after that position.

Please check the screenshot attached here. It's entirely possible that we could name a folder with "=" in it. So, this solution might break for those cases.

screen shot 2018-05-09 at 10 30 11 am

VishwasShashidhar:
Fixed the special character use case in the path.

PR #364, opened 2018-05-10
ELECTRON-481: fix the error dialogs with always on top use case

VishwasShashidhar:
## Description
When always on top is enabled and an error dialog is to be shown, the dialog goes in the background and the menu items get disabled on Windows. To fix this, we use a message box with the type "error" instead of an error box [ELECTRON-481](https://perzoinc.atlassian.net/browse/ELECTRON-481)

## Approach
- #### Problem with the code: Using `showErrorBox` to show dialogs
- #### Fix: Using `showMessageBox` to show dialogs with type "error"

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

- [x] Unit-Tests
[ELECTRON-481 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1991003/ELECTRON-481.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-481 Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/1991077/ELECTRON-481.Spectron.Tests.pdf)

### DIFF
--- a/js/downloadManager/index.js
+++ b/js/downloadManager/index.js
@@ -27,8 +27,9 @@ function openFile(id) {
     });
     if (fileIndex !== -1) {
         let openResponse = remote.shell.openExternal(`file:///${local.downloadItems[fileIndex].savedPath}`);
-        if (!openResponse) {
-            remote.dialog.showErrorBox("File not found", 'The file you are trying to open cannot be found in the specified path.');
+        let focusedWindow = remote.BrowserWindow.getFocusedWindow();
+        if (!openResponse && focusedWindow && !focusedWindow.isDestroyed()) {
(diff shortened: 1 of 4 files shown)


VikasShashidhar:
@VishwasShashidhar Are there going to be any design or UX differences compared to when using `showErrorBox`? If so, we'll need to talk to design and product. Maybe we can use a separate ticket top track that, but this is worth noting. I am approving this PR though.

VishwasShashidhar:
@VikasShashidhar Yes, on macOS, it'll be different, not really blocker. But, I've planned to talk to Savvi and take it up in the next Sprint.

PR #368, opened 2018-05-15
ELECTRON-491 (Fixes a Javascript error when pop-in a meeting)

KiranNiranjan:
## Description
Change method to use getFocusedWindow instead of fromWebContents [ELECTRON-491](https://perzoinc.atlassian.net/browse/ELECTRON-491)

## Approach
#### Problem with the code:
 - Calling a `fromWebContents` method with undefined `webContents` throwing an error
 
tng01_1

#### Fix:
 - Using `getFocusedWindow` for get the browser window access
 - Cleaned up spaces

## Spectron tests result
[Electron-491 — Spectron.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2004714/Electron-491.Spectron.pdf)

## Unit tests result
[Electron-491 — Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2004713/Electron-491.Unit.Tests.pdf)

- [X] Unit-Tests
- [X] Documentation
- [X] Automation-Tests

### DIFF
--- a/js/windowMgr.js
+++ b/js/windowMgr.js
@@ -496,32 +496,35 @@ function doCreateMainWindow(initialUrl, initialBounds) {
      * @param webContents Web contents of the window
      */
     function handlePermissionRequests(webContents) {
-        
+
         let session = webContents.session;
-        
+
         getConfigField('permissions')
             .then((permissions) => {
-                
+
                 if (!permissions) {
                     log.send(logLevels.ERROR, 'permissions configuration is invalid, so, everything will be true by default!');
                     return;
                 }
-                
+
                 session.setPermissionRequestHandler((sessionWebContents, permission, callback) => {
-                    
+
                     function handleSessionPermissions(userPermission, message, cb) {
-                        
+
                         log.send(logLevels.INFO, 'permission is -> ' + userPermission);
-                        
+
                         if (!userPermission) {
                             let fullMessage = `Your administrator has disabled ${message}. Please contact your admin for help.`;
-                            electron.dialog.showMessageBox(BrowserWindow.fromWebContents(webContents), {type: 'error', title: 'Permission Denied!', message: fullMessage});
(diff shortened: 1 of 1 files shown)


VishwasShashidhar:
@VikasShashidhar Please review this.

js/windowMgr.js
```diff
@@ -496,73 +496,76 @@ function doCreateMainWindow(initialUrl, initialBounds) {
...
+
                         log.send(logLevels.INFO, 'permission is -> ' + userPermission);
-                        
+
                         if (!userPermission) {
                             let fullMessage = `Your administrator has disabled ${message}. Please contact your admin for help.`;
-                            electron.dialog.showMessageBox(BrowserWindow.fromWebContents(webContents), {type: 'error', title: 'Permission Denied!', message: fullMessage});
+                            const browserWindow = BrowserWindow.getFocusedWindow();
+                            if (browserWindow && !browserWindow.isDestroyed()) {
+                                electron.dialog.showMessageBox(browserWindow, {type: 'error', title: 'Permission Denied!', message: fullMessage});
```
VikasShashidhar: So, if `browserWindow` is not accessible/available, does that mean we don't show the error message at all?

KiranNiranjan: Yes that's correct, We don't show the error message if the `browserWindow` is not available

VikasShashidhar: Ok. But, how are we going to communicate to the user that that particular functionality is unavailable to him/her?

KiranNiranjan: @VikasShashidhar This only happens when the user starts the meeting and then changes the focus on to a different application, but there is a red banner shown for a few secs 
screen shot 2018-05-21 at 1 49 22 pm

The electron dialog will be shown to the user when he/she interacts with any of the actions buttons (In this case the `browserWindow` will always be defined)

PR #375, opened 2018-05-21
ELECTRON-479: fix launch on startup issue on first time launch

VishwasShashidhar:
## Description
Fix launch on startup issues when the app is installed and launched first time [ELECTRON-479](https://perzoinc.atlassian.net/browse/ELECTRON-479)

## Approach
- #### Problem with the code: String to boolean comparison doesn't work in the method setStartup of main.js
- #### Fix: Fixed the comparison to use appropriate logic.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

- [x] Unit-Tests
[ELECTRON-479 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2023109/ELECTRON-479.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-479 Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2023129/ELECTRON-479.Spectron.Tests.pdf)

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -303,7 +303,7 @@ function setupFirstTimeLaunch() {
 function setStartup(lStartup) {
     log.send(logLevels.INFO, `launch on startup parameter value is ${lStartup}`);
     return new Promise((resolve) => {
-        let launchOnStartup = JSON.parse(lStartup);
+        let launchOnStartup = (String(lStartup) === 'true');
         log.send(logLevels.INFO, `launchOnStartup value is ${launchOnStartup}`);
         if (launchOnStartup) {
             log.send(logLevels.INFO, `enabling launch on startup`);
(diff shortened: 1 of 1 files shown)


js/main.js
```diff
@@ -301,9 +301,12 @@ function setupFirstTimeLaunch() {
  * @returns {Promise}
  */
 function setStartup(lStartup) {
+    log.send(logLevels.INFO, `launch on startup parameter value is ${lStartup}`);
     return new Promise((resolve) => {
-        let launchOnStartup = (lStartup === 'true');
+        let launchOnStartup = (String(lStartup) === 'true');
```
VikasShashidhar: Why not use the `toString()` method?

VishwasShashidhar: `String(value)` works properly in case of `null` / `undefined` cases whereas `toString()` throws error.

PR #385, opened 2018-05-29
Non cases sensitive cmd line args

danielgronberg:
## Description
Updated handling of cmd line args to be case insensitive.

### DIFF
--- a/js/utils/getCmdLineArg.js
+++ b/js/utils/getCmdLineArg.js
@@ -4,7 +4,7 @@ const log = require('../log.js');
 const logLevels = require('../enums/logLevels.js');
 
 /**
- * Search given argv for argName using exact match or starts with.
+ * Search given argv for argName using exact match or starts with. Comparison is case insensitive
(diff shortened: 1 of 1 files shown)


js/utils/getCmdLineArg.js
```diff
@@ -9,17 +9,20 @@ const logLevels = require('../enums/logLevels.js');
  * @param  {String} argName   Arg name to search for.
  * @param  {Boolean} exactMatch  If true then look for exact match otherwise
  * try finding arg that starts with argName.
+ * @param {Boolean} caseSensitive Optional. If set to false, the comparison is not case sensitive
  * @return {String}           If found, returns the arg, otherwise null.
  */
-function getCmdLineArg(argv, argName, exactMatch) {
+function getCmdLineArg(argv, argName, exactMatch, caseSensitive=true) {
```
danielgronberg: Alternatively, we don't have an arg and simply always do cases insensitive comparison. Either way works for me

PR #392, opened 2018-06-07
SEARCH-729 (New Archive Index on Disk)

keerthi16:
## Description
Archive Index on Disk
[SEARCH-729](https://perzoinc.atlassian.net/browse/SEARCH-729)

This PR also covers:
[SEARCH-765](https://perzoinc.atlassian.net/browse/SEARCH-765)
[SEARCH-766](https://perzoinc.atlassian.net/browse/SEARCH-766)
[SEARCH-770](https://perzoinc.atlassian.net/browse/SEARCH-770)
[SEARCH-764](https://perzoinc.atlassian.net/browse/SEARCH-764)

Document of the implementation can be found here:
https://perzoinc.atlassian.net/wiki/spaces/SEARCH/pages/144711439/Client+Side+Search+-+Electron+Integration

## Approach
- #### Problem with the code: Previous implementation which was storing plane message on the disk and then it was encrypting.
- #### Fix: New implementation receives encrypted messages from the library which then archived and stored on the disk.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

## Unit Tests Result
[Test Results — Unit-Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2080868/Test.Results.Unit-Tests.pdf)

## Spectron Tests Result
[Test Results — Spectron.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2080871/Test.Results.Spectron.pdf)

- [x] Unit-Tests
- [x] Documentation
- [x] Automation-Tests


js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
...
+                let indexDateStartFrom = new Date().getTime() - searchConfig.SEARCH_PERIOD_SUBTRACTOR;
+                // Deleting all the messages except 3 Months from now
+                libSymphonySearch.symSEDeleteMessagesFromRAMIndex(null,
+                    searchConfig.MINIMUM_DATE, indexDateStartFrom.toString());
+            });
+        }
+    }
+
+    deCompress(key) {
+        let userIndexPath = path.join(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH,
```
KiranNiranjan: same as above

js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
...
+            log.send(logLevels.ERROR, 'Batch Indexing: Messages not provided');
+            return callback(false, 'Batch Indexing: Messages are required');
+        }
 
-            if (!fs.existsSync(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH)) {
-                log.send(logLevels.ERROR, 'User index folder not found');
-                reject(new Error('User index folder not found'));
-                return;
+        try {
+            let msg = JSON.parse(messages);
```
VikasShashidhar: You can probably rename this to something like `messagesData`. `msg` gives an impression that it's an individual message and not a set of them.

js/search/search.js
```diff
@@ -178,57 +158,72 @@ class Search {
      * An array of messages to be indexed
      * in real time
      * @param message
+     * @param callback
      */
-    realTimeIndexing(message) {
+    realTimeIndexing(message, callback) {
```
VikasShashidhar: Rename to `messages` please.

js/search/search.js
```diff
@@ -178,57 +158,72 @@ class Search {
...
+        if (typeof callback !== "function") {
+            return false;
+        }
 
         try {
             let msg = JSON.parse(message);
```
VikasShashidhar: Same as above.

js/search/search.js
```diff
@@ -178,57 +158,72 @@ class Search {
...
             if (!(msg instanceof Array)) {
-                log.send(logLevels.ERROR, 'RealTime Indexing: Messages must be an array real-time indexing');
-                return (new Error('RealTime Indexing: Messages must be an array'));
+                return callback(false, 'RealTime Indexing: Messages must be an array');
```
VikasShashidhar: Can we abstract out all this logic to a separate method? Let's avoid redundancy.

js/search/search.js
```diff
@@ -178,57 +158,72 @@ class Search {
...
+        return null;
     }
 
     /**
      * Encrypting the index after the merging the index
      * to the main user index
      */
     encryptIndex(key) {
-        return this.crypto.encryption(key);
+        const mainIndexFolder = path.join(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH,
```
VikasShashidhar: We could abstract out the entire path validation so that many methods can make use of it.

js/search/search.js
```diff
@@ -178,57 +158,72 @@ class Search {
...
+            `${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME}_${this.userId}`);
+        return new Promise(resolve => {
+            if (!fs.existsSync(mainIndexFolder)) {
```
VikasShashidhar: What we could do here instead is have a method that could check if the `mainIndexFolder` exists during initialisation and set a bool in this class so that all these methods could use it.

Or, if you believe that a user or any other application/program could arbitrarily remove these folders, then make a check every time we use this method by invoking that abstracted out method.

js/search/search.js
```diff
@@ -519,160 +517,33 @@ class Search {
...
-     * Removing all the folders and files inside the data folder
-     * @param location
-     */
-    static deleteIndexFolders(location) {
-        if (fs.existsSync(location)) {
-            fs.readdirSync(location).forEach((file) => {
-                let curPath = location + "/" + file;
+function clearSearchData() {
+    function removeFiles(filePath) {
+        if (fs.existsSync(filePath)) {
```
VikasShashidhar: Same as above.

js/search/search.js
```diff
@@ -519,160 +517,33 @@ class Search {
...
-                if (response) {
-                    log.send(logLevels.INFO, 'Launch Agent: Creating successful');
-                }
-            });
-        });
-    } else {
-        let folderPath = isDevEnv ? path.join(__dirname, '..', '..', searchConfig.FOLDERS_CONSTANTS.INDEX_FOLDER_NAME) :
-            path.join(searchConfig.FOLDERS_CONSTANTS.USER_DATA_PATH, searchConfig.FOLDERS_CONSTANTS.INDEX_FOLDER_NAME);
-        taskScheduler(`${searchConfig.LIBRARY_CONSTANTS.WINDOWS_TASK_FILE}`, folderPath, pidValue, `${searchConfig.LIBRARY_CONSTANTS.WINDOWS_CLEAR_SCRIPT}`);
+    if (this.userId) {
```
VikasShashidhar: Another potential place of exploitation. We need to add more validations if possible.

js/search/searchLibrary.js
```diff
@@ -13,6 +13,25 @@ const symLucyIndexerPtr = ref.refType(symLucyIndexer);
  * using the node-ffi
  */
 let libSymphonySearch = ffi.Library(searchConfig.LIBRARY_CONSTANTS.SEARCH_LIBRARY_PATH, {
+
+    //New Memory Indexing API
```
VikasShashidhar: Should be `// New Memory Indexing API`. Please mind the linting before push a commit.

js/preload/preloadMain.js
```diff
@@ -92,7 +92,8 @@ function createAPI() {
                     containerIdentifier: appName,
                     containerVer: appVer,
                     buildNumber: buildNumber,
-                    apiVer: '1.0.0'
+                    apiVer: '2.0.0',
```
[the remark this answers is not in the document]
keerthi16: @VikasShashidhar Previously I was using `apiVer` now I created `searchApiVer`

js/search/search.js
```diff
@@ -1,22 +1,17 @@
...
+const ref = require('ref');
 const path = require('path');
-const isDevEnv = require('../utils/misc.js').isDevEnv;
-const isMac = require('../utils/misc.js').isMac;
 const makeBoundTimedCollector = require('./queue');
 const searchConfig = require('./searchConfig');
 const log = require('../log.js');
 const logLevels = require('../enums/logLevels.js');
-const { launchAgent, launchDaemon, taskScheduler } = require('./utils/search-launchd.js');
+const lz4 = require('../compressionLib');
```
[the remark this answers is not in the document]
keerthi16: Sure 👍

js/preload/preloadMain.js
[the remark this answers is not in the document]
VikasShashidhar: Ok. So, might we need this for backward compatibility then?

js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
...
+                let indexDateStartFrom = new Date().getTime() - searchConfig.SEARCH_PERIOD_SUBTRACTOR;
+                // Deleting all the messages except 3 Months from now
+                libSymphonySearch.symSEDeleteMessagesFromRAMIndex(null,
+                    searchConfig.MINIMUM_DATE, indexDateStartFrom.toString());
+            });
+        }
+    }
+
+    deCompress(key) {
+        const userIndexPath = path.join(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH,
```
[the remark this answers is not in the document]
keerthi16: We need the user id which is only available in search class and if we abstract this we will expose user system path

js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
...
+            `${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME}_${this.userId}`);
+        if (fs.existsSync(`${userIndexPath}${searchConfig.TAR_LZ4_EXT}`)) {
+            lz4.deCompression(`${userIndexPath}${searchConfig.TAR_LZ4_EXT}`, (err) => {
+                if (err && !fs.existsSync(userIndexPath)) {
```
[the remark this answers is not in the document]
keerthi16: No, this is just checking if the file `search_index_123456.tar.lz4` exist or not and the userIndexPath is only `search_index_123456`with out ext which will never exist in this method.

js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
...
+                    fs.mkdirSync(userIndexPath);
+                }
+                this.init(key);
+            })
+        } else {
+            if (!fs.existsSync(userIndexPath)) {
+                fs.mkdirSync(userIndexPath);
```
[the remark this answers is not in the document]
keerthi16: Yes

js/search/search.js
keerthi16: As the Search class is exposed we cannot pass path as an argument which exposes user system path.

js/preload/preloadMain.js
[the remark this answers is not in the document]
keerthi16: Yes

js/search/search.js
```diff
@@ -314,42 +314,41 @@ class Search {
...
+
+        return null;
     }
 
-    /*eslint class-methods-use-this: ["error", { "exceptMethods": ["deleteRealTimeFolder"] }] */
+    /**
+     * This function clears the real-time index
+     * before starting the batch-indexing
+     */
     deleteRealTimeFolder() {
```
VishwasShashidhar: should this be refactored appropriately?

js/search/search.js
```diff
@@ -519,160 +518,61 @@ class Search {
...
+                let curPath = filePath + "/" + file;
                 if (fs.lstatSync(curPath).isDirectory()) {
-                    Search.deleteIndexFolders(curPath);
+                    removeFiles(curPath);
                 } else {
                     fs.unlinkSync(curPath);
                 }
             });
-            fs.rmdirSync(location);
+            fs.rmdirSync(filePath);
```
VishwasShashidhar: won't work if there are files inside the dir

js/preload/preloadMain.js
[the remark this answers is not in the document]
keerthi16: @VishwasShashidhar The `apiVer` was not introduced as part search. I just changed the version to `2.0.0` as I removed the some of the API from the preload script. 
Also, I added `searchApiVer` for backward compatibility.

js/search/search.js
```diff
@@ -66,94 +46,95 @@ class Search {
      * initialise the SymphonySearchEngine library
      * and creates a folder in the userData
      */
-    init() {
+    init(key) {
+        let bufKey = new Buffer(key, searchConfig.KEY_ENCODING);
+        if (!key && bufKey.length !== searchConfig.KEY_LENGTH) {
+            return;
+        }
+        libSymphonySearch.symSEDestroy();
```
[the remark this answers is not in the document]
keerthi16: All this method are from the search library.

js/search/search.js
```diff
@@ -66,94 +46,95 @@ class Search {
...
-        libSymphonySearch.symSEEnsureFolderExists(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH);
-        Search.deleteIndexFolders(searchConfig.FOLDERS_CONSTANTS.TEMP_REAL_TIME_INDEX);
-        Search.deleteIndexFolders(searchConfig.FOLDERS_CONSTANTS.TEMP_BATCH_INDEX_FOLDER);
-        Search.indexValidator(`${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME_PATH}_${this.userId}`);
-        Search.indexValidator(searchConfig.FOLDERS_CONSTANTS.TEMP_REAL_TIME_INDEX);
-        let indexDateStartFrom = new Date().getTime() - searchConfig.SEARCH_PERIOD_SUBTRACTOR;
-        // Deleting all the messages except 3 Months from now
-        libSymphonySearch.symSEDeleteMessages(`${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME_PATH}_${this.userId}`, null,
-            searchConfig.MINIMUM_DATE, indexDateStartFrom.toString());
+        libSymphonySearch.symSEClearMainRAMIndex();
```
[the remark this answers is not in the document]
keerthi16: All this method are from the search library.

js/search/search.js
keerthi16: screen shot 2018-06-08 at 8 18 54 pm

All the files inside the folder will be deleted before deleting the folder.

js/search/queue.js
```diff
@@ -20,12 +23,21 @@ let makeBoundTimedCollector = function(isIndexing, timeout, callback) {
         }
     };
 
+    function handleRealTimeResponse(status, response) {
```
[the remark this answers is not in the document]
keerthi16: The function requires a callback but we don't need to take any actions so just logging the state.

js/search/search.js
```diff
@@ -540,6 +544,34 @@ function clearSearchData() {
...
+ */
+function isFileExist(type) {
+    let searchPath;
+
+    if (!this.userId) {
+        return false
+    }
+
+    switch (type) {
+        case 'USER_INDEX_PATH':
```
VikasShashidhar: Instead of using `switch case...` here, what we could do is keep all these values in an object and based on `type`'s value, we could return the corresponding value from the object.

Just a suggestion. That way, it's easier to grow your cases in the future rather than having long `switch case...` statements.

js/search/searchConfig.js
```diff
@@ -58,7 +58,9 @@ const searchConfig = {
     DISK_NOT_READY: 'NOT_READY',
     DISK_NOT_FOUND: 'DISK_NOT_FOUND',
     MINIMUM_DISK_SPACE: 300000000, // in bytes
-    MAC_PATH_ERROR: "No such file or directory"
+    MAC_PATH_ERROR: "No such file or directory",
```
VikasShashidhar: Can you please talk to @VishwasShashidhar and localize this? We can avoid doing this in a separate ticket.

js/search/search.js
```diff
@@ -556,18 +556,14 @@ function isFileExist(type) {
...
-            break;
-    }
+    const pathObject = {
+        USER_INDEX_PATH: path.join(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH,
+            `${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME}_${this.userId}`),
+        LZ4: path.join(searchConfig.FOLDERS_CONSTANTS.INDEX_PATH,
+            `${searchConfig.FOLDERS_CONSTANTS.PREFIX_NAME}_${this.userId}${searchConfig.TAR_LZ4_EXT}`),
+    };
+
+    searchPath = pathObject[type];
```
VikasShashidhar: Would be better to check if `type` exists in `pathObject` and then assign it. Also, please rename `pathObject` to something better.

keerthi16: If `type` is undefined the `searchPath` will be undefined and I already have a check for that.

Will rename the `pathObject` to paths

js/search/search.js
```diff
@@ -66,94 +46,94 @@ class Search {
      * initialise the SymphonySearchEngine library
      * and creates a folder in the userData
      */
-    init() {
+    init(key) {
+        if (!key) {
```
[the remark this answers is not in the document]
VikasShashidhar: @keerthi16 Once this is addressed, I'll approve this PR.

js/search/search.js
VikasShashidhar: @keerthi16 Once this is addressed, I'll approve this PR.

js/search/search.js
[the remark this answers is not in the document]
keerthi16: @VikasShashidhar This is already addressed the code change is in the next line

js/search/search.js
keerthi16: I'm not really sure how we need to handle this as this is just a random number, and this is just used to name the index file.

PR #401, opened 2018-06-18
ELECTRON-519: support Japanese localization for menus, dialogs and windows

VishwasShashidhar:
## Description
Support Japanese localization for the main menu and it's items, screen picker window, notification configuration window, authentication dialog, error and info dialogs. 
[ELECTRON-519](https://perzoinc.atlassian.net/browse/ELECTRON-519)
[ELECTRON-536](https://perzoinc.atlassian.net/browse/ELECTRON-536)

## Approach
- When the app starts, check to see if we have locale set in the user config
- If the locale is set, pick the locale from user config and use that. If the locale is not set, use the system default locale using the app.getLocale() function provided by Electron
- Use the above locale and build the menu items and use the corresponding strings from the translation files across the app for the appropriate places mentioned in the description.
- When the language is changed on the web app, send an event to electron and update the main menu items and the user config.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

- [x] Unit-Tests
[ELECTRON-519 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2110274/ELECTRON-519.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-519 Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2110275/ELECTRON-519.Spectron.Tests.pdf)


js/mainApiMgr.js
```diff
@@ -143,7 +143,7 @@ electron.ipcMain.on(apiName, (event, arg) => {
             }
             break;
         case apiCmds.popupMenu:
-            var browserWin = electron.BrowserWindow.fromWebContents(event.sender);
+            let browserWin = electron.BrowserWindow.fromWebContents(event.sender);
```
KiranNiranjan: eslint will fail here. This Block of code shoud go inside `{}`
``` javascript
        case apiCmds.popupMenu: {
            let browserWin = electron.BrowserWindow.fromWebContents(event.sender);
            if (browserWin && !browserWin.isDestroyed()) {
                windowMgr.getMenu().popup(browserWin, {x: 20, y: 15, async: true});
            }
            break;
        }
```

PR #407, opened 2018-06-20
ELECTRON-535 (Add support for screen snippet localization)

KiranNiranjan:
## Description
Bump screen snippet version to 1.0.2 and add support for screen snippet localization [ELECTRON-535](https://perzoinc.atlassian.net/browse/ELECTRON-535)

## Approach
#### Fix:
 - Bump screen snippet version to 1.0.2
 - Add support for screen snippet localization by passing locale arg as the second arg

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
ScreenSnippet | [link](https://github.com/symphonyoss/ScreenSnippet/pull/4)

- [ ] Unit-Tests
- [ ] Documentation
- [ ] Automation-Tests

### DIFF
--- a/js/translation/i18n.js
+++ b/js/translation/i18n.js
@@ -16,7 +16,16 @@ const setLanguage = function(lng) {
     loadedTranslations = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'locale', language + '.json'), 'utf8'));
 };
 
+/**
+ * Returns the current locale
+ * @return {*|string}
+ */
+const getLanguage = function() {
+    return language || 'en-US';
+};
+
 module.exports = {
     setLanguage: setLanguage,
-    getMessageFor: getMessageFor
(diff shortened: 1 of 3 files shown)


js/translation/i18n.js
```diff
@@ -16,7 +16,16 @@ const setLanguage = function(lng) {
     loadedTranslations = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'locale', language + '.json'), 'utf8'));
 };
 
+/**
+ * Returns the current locale
+ * @return {*|string}
+ */
+const getLanguage = function() {
+    return language || 'en-US';
```
VikasShashidhar: Can't we default it to the system locale setting instead?

VishwasShashidhar: the `language` variable is set through either user config / system locale. This is a safety check in case `language` is undefined, we fallback to the default language i.e. `en-US`

PR #421, opened 2018-07-09
ELECTRON-512 (Create API only if the location.origin matches with whitelist or pod URL)

KiranNiranjan:
## Description
Safety to protect from exposing `window.ssf` API to an unknown origin [ELECTRON-512](https://perzoinc.atlassian.net/browse/ELECTRON-512)

## Solution Approach
Emitting origin check event to the main process whenever the preload script is loaded
The main process receives the event published by the renderer process and validates the origin 
if the origin is valid the main process emits an event to the renderer process which then creates the API

## Documentation
N/A

## QA Checklist
- [X] Unit-Tests
[ELECTRON-512 — Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2174816/ELECTRON-512.Unit.Tests.pdf)

- [X] Automation-Tests
[Electron-512 — Spectron.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2174818/Electron-512.Spectron.pdf)

### DIFF
--- a/js/originCheck.js
+++ b/js/originCheck.js
@@ -1,0 +1,33 @@
+const { isWhitelisted, matchDomains, parseDomain } = require('./utils/whitelistHandler');
+const { getGlobalConfigField } = require('./config');
+const { isDevEnv } = require('./utils/misc');
+
+/**
+ * Validate whitelist and location.origin
+ * @param eventSender
+ * @param origin {String} location.origin
+ * @return {Promise<T>}
+ */
+function originCheck(eventSender, origin) {
+
+    if (isDevEnv) {
+        eventSender.send('initialize-api');
+        return null;
+    }
+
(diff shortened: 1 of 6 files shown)


KiranNiranjan:
Added `do-not-merge` until we test this functionality completely

js/originCheck.js
```diff
@@ -0,0 +1,33 @@
...
+ * @return {Promise<T>}
+ */
+function originCheck(eventSender, origin) {
+
+    if (isDevEnv || isNodeEnv) {
+        eventSender.send('initialize-api');
+        return null;
+    }
+
+    return isWhitelisted(origin)
```
VishwasShashidhar: do we need a `return` here?

KiranNiranjan: Nope. Removed

PR #481, opened 2018-08-28
ELECTRON-602: fix the pre-install script on macOS installation / upgrade cases

VishwasShashidhar:
## Description
Currently, when we try to install a version that is lower than the currently installed version, the pre install script deletes the existing version and the new version is not installed either. This is because of how macOS handles upgrades. 
[ELECTRON-602](https://perzoinc.atlassian.net/browse/ELECTRON-602)

## Solution Approach
So, we add a version check function that exits the script if we have a new version already installed on the system rather than deleting the existing version installed on the system

## Documentation
N/A

## Related PRs
N/A

## QA Checklist
- [x] Unit-Tests
[ELECTRON-602 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2327259/ELECTRON-602.Unit.Tests.pdf)

### DIFF
--- a/installer/mac/preinstall.sh
+++ b/installer/mac/preinstall.sh
@@ -1,3 +1,60 @@
 #!/usr/bin/env bash
+
+# Kill the existing running instance
 sudo killall Symphony
-sudo rm -rf /Applications/Symphony.app
+
+compare_versions()
+{
+    # Get the installer version:
+    CURRENT_VERSION=3.1.0
+
+    # Get the currently installed version:
+    INSTALLED_VERSION=$(plutil -p /Applications/Symphony.app/Contents/Info.plist | awk '/CFBundleShortVersionString/ {print substr($3, 2, length($3)-2)}')
+
+    # If there are no versions installed, just exit the script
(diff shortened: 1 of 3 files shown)


installer/mac/preinstall.sh
```diff
@@ -1,3 +1,60 @@
...
+    minor2=$(echo ${INSTALLED_VERSION} | awk '{print $2}')
+    major2=$(echo ${INSTALLED_VERSION} | awk '{print $1}')
+
+    # And now, we can simply compare the variables:
+    if [ ${major1} -lt ${major2} ]; then
+        echo "Installed version is newer than this version, exiting installation"
+        exit 1
+    fi
+
+    if [ ${minor1} -lt ${minor2} ]; then
```
KiranNiranjan: The installation would fail if the older version is `2.11.8` and the new version is `3.1.0`, as this is only comparing the minor versions

VishwasShashidhar: Yes indeed, noticed that 👍 will fix and push the code.

PR #484, opened 2018-08-30
ELECTRON-635 (Custom auto launch implementation)[Windows]

KiranNiranjan:
## Description
Add config field where admin can specific a custom `exe` path for which the auto launch is set up [ELECTRON-635](https://perzoinc.atlassian.net/browse/ELECTRON-635)

- Optimize auto launch
- Optimize first time launch config functionality

## Solution Approach
Checks if the config field `autoLaunchPath` exists, if exists creates a launchagent/register entries with the specified path or falls back to the `Symphony app path`

## QA Checklist
- [X] Unit-Tests
[ELECTRON-635 Unit Test Result.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2334714/ELECTRON-635.Unit.Test.Result.pdf)

- [ ] Automation-Tests


js/config.js
```diff
@@ -258,14 +256,18 @@ function updateUserConfig(oldUserConfig) {
...
 
         // write the new user config changes to the user config file
-        fs.writeFileSync(userConfigFile, newUserConfigString, 'utf-8');
-
-        return resolve();
-
+        fs.writeFile(userConfigFile, newUserConfigString, 'utf-8', (err) => {
+            if (err) {
+                reject(`Failed to update user config error: ${err}`);
+                return;
```
VishwasShashidhar: this can be moved to the above line?

KiranNiranjan: Then we would also need to `return fs.writeFile` which I think is not a good idea

PR #491, opened 2018-09-03
ELECTRON-710 - Fix Always on top: notification does not show on top

vrbsm:
## Description
Electron - Always on top - Toast notification does not show on top (front) of the Electron if close toast notification before

## Fix
electron-710

## QA Checklist
- [x] Unit-Tests
- [ ] Automation-Tests

Attach unit & spectron tests results in PDF format against the above task lists for this branch

### DIFF
--- a/js/notify/electron-notify.js
+++ b/js/notify/electron-notify.js
@@ -422,6 +422,14 @@ function setNotificationContents(notfWindow, notfObj) {
     // Display time per notification basis.
     let displayTime = notfObj.displayTime ? notfObj.displayTime : config.displayTime;
 
+    let browserWindows = BrowserWindow.getAllWindows();
+    const mainWindow = browserWindows.find((window) => { return window.winName === 'main' });
+    if (mainWindow && !mainWindow.isDestroyed()) {
+        if (mainWindow.isAlwaysOnTop()) {
+            notfWindow.setAlwaysOnTop(true);
+        }
+    }
+
     if (notfWindow.displayTimer) {
         clearTimeout(notfWindow.displayTimer);
     }


js/notify/electron-notify.js
```diff
@@ -421,7 +421,7 @@ function setNotificationContents(notfWindow, notfObj) {
 
     // Display time per notification basis.
     let displayTime = notfObj.displayTime ? notfObj.displayTime : config.displayTime;
-
+    notfWindow.setAlwaysOnTop(true);
```
VishwasShashidhar: this should be selective and should only be enabled if this user has "Always on Top" enabled on our app. If not, it'll be always on top even if our app is not in focus.

vrbsm: @VishwasShashidhar ok, I fixed it.
Thanks

js/notify/electron-notify.js
```diff
@@ -422,6 +423,14 @@ function setNotificationContents(notfWindow, notfObj) {
     // Display time per notification basis.
     let displayTime = notfObj.displayTime ? notfObj.displayTime : config.displayTime;
 
+    getConfigField('alwaysOnTop').then((isAlwaysOnTop) => {
+        if (isAlwaysOnTop) {
```
VishwasShashidhar: this is asynchronous and by the time we get the data from config, this process would've finished and it won't work.

please have a local variable for alwaysOnTop, read that during the initialisation process, update it when the setting changes and use that variable to make this check.

VishwasShashidhar:
I'll merge this once we have the new build ready.

PR #511, opened 2018-09-29
ELECTRON-801 (Dynamically update notification based on message length)

KiranNiranjan:
## Description
Dynamically update notification based on message length [ELECTRON-801](https://perzoinc.atlassian.net/browse/ELECTRON-801)

## Solution Approach
Resize/Rearrange notification content based on the message text length

## Screen shot
screenshot 2018-09-29 at 7 08 15 pm

## QA Checklist
- [X] Unit-Tests
[Electron-801 Test Result.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2430660/Electron-801.Test.Result.pdf)

- [ ] Automation-Tests

### DIFF
--- a/js/notify/electron-notify.html
+++ b/js/notify/electron-notify.html
@@ -2,7 +2,7 @@
 <head></head>
 <body style='margin:0; overflow: hidden; -webkit-user-select: none;'>
 <div id="container">
-    <div>
+    <div id="logo-container">
         <img src="" id="symphony-logo">
     </div>
     <div id="header">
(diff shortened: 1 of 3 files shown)


js/notify/electron-notify.html
```diff
@@ -2,7 +2,7 @@
 <head></head>
 <body style='margin:0; overflow: hidden; -webkit-user-select: none;'>
 <div id="container">
-    <div>
+    <div id="logo-container">
         <img src="" id="symphony-logo">
```
VikasShashidhar: Is it not better to have a default image in place?

PR #535, opened 2018-12-12
ELECTRON-956 - removing the minimize and maximize from the configure desktop alert position

vrbsm:
## Description
App is not minimized upon "Configure Desktop Alert Position" modal when "Always on Top" = True.  [ELECTRON-956](https://perzoinc.atlassian.net/browse/ELECTRON-956)

## Fix
electron

## QA Checklist
- [x] Unit-Tests
- [ ] Automation-Tests

Attach unit & spectron tests results in PDF format against the above task lists for this branch
[UnitTest.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2673191/UnitTest.pdf)

### DIFF
--- a/js/windowMgr.js
+++ b/js/windowMgr.js
@@ -504,6 +504,7 @@ function doCreateMainWindow(initialUrl, initialBounds, isCustomTitleBar) {
                     newWinOptions.minHeight = MIN_HEIGHT;
                     newWinOptions.alwaysOnTop = alwaysOnTop;
                     newWinOptions.frame = true;
+                    newWinOptions.parent = null;
 
                     let newWinKey = getGuid();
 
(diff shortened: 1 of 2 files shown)


js/stats.js
```diff
@@ -44,6 +44,19 @@ const logPodStats = () => {
         });
 };
 
+/**
+ * ELECTRON-956: App is not minimized upon "Configure Desktop Alert Position" modal when "Always on Top" = True
+ */
+app.on('browser-window-created', (event, window) => {
```
VishwasShashidhar: move this logic to window manager file. Doesn't make sense to have it in stats.js

js/windowMgr.js
```diff
@@ -741,6 +741,19 @@ function doCreateMainWindow(initialUrl, initialBounds, isCustomTitleBar) {
 
 }
 
+/**
+ * ELECTRON-956: App is not minimized upon "Configure Desktop Alert Position" modal when "Always on Top" = True
+ */
+app.on('browser-window-created', (event, window) => {
+    const parentWindow = window.getParentWindow();
+    if (parentWindow && !parentWindow.isDestroyed()) {
+        if (parentWindow.winName === 'main') {
```
VishwasShashidhar: @KiranNiranjan will this work in case of email child pop-out windows when there is more than 1 level?

KiranNiranjan: @VishwasShashidhar This will work for more than 1 level. But this functionality will also disable `minimizable` and `maximizable` controls for the chat pop-outs

vrbsm: Yes, if there is a parent associated with pop-out

VishwasShashidhar: @KiranNiranjan @vrbsm If this disables `minimize` & `maximize` for our regular pop-outs, that would be a problem. We only need to apply this to all the special case windows like "About Symphony", "Notification Preferences" and so on.

VishwasShashidhar:
@vrbsm Please see comments and address them so I can merge the PR.

PR #547, opened 2019-01-14
ELECTRON-869 (Enable spellchecker for main window)

KiranNiranjan:
## Description
Enables spellchecker for main window [link](https://perzoinc.atlassian.net/browse/ELECTRON-869)

## Solution Approach
Enable spellchecker only for the main window until the underline memory leak is fixed on electron
https://github.com/electron/electron/issues/15459

## Screen shot
screenshot 2019-01-14 at 4 29 48 pm

## Related PRs
List related PRs against other branches / repositories:

branch | PR
------ | ------
Electron | [Issue](https://github.com/electron/electron/issues/15459)

## QA Checklist
- [X] Unit-Tests
[Symphony Electron Test Result.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2754871/Symphony.Electron.Test.Result.pdf)

- [ ] Automation-Tests

### DIFF
--- a/js/preload/preloadMain.js
+++ b/js/preload/preloadMain.js
@@ -11,7 +11,7 @@
 // also to bring pieces of node.js:
 // https://github.com/electron/electron/issues/2984
 //
-const { ipcRenderer, remote, crashReporter } = require('electron');
+const { ipcRenderer, remote, crashReporter, webFrame } = require('electron');
 
 const throttle = require('../utils/throttle.js');
 const apiEnums = require('../enums/api.js');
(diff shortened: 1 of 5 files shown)


js/preload/preloadMain.js
```diff
@@ -86,6 +86,20 @@ const throttledSetIsInMeetingStatus = throttle(1000, function (isInMeeting) {
 local.ipcRenderer.on('on-page-load', () => {
     snackBar = new SnackBar();
 
+    // Enable spellchecker only for main window until
+    // the underline memory leak is fixed on electron
+    // https://github.com/electron/electron/issues/15459
+    if (window.name === 'main') {
```
VishwasShashidhar: is there a better way to check if it's the main window? Maybe using the parent property or something like that?

js/preload/preloadMain.js
```diff
@@ -86,6 +86,20 @@ const throttledSetIsInMeetingStatus = throttle(1000, function (isInMeeting) {
...
+        webFrame.setSpellCheckProvider('en-US', true, {
+            spellCheck (text) {
+                return !local.ipcRenderer.sendSync(apiName, {
```
VishwasShashidhar: anyway we can change this async to avoid blocking of the main thread?

KiranNiranjan: I don't think this is possible yet. https://github.com/electron/electron/issues/2065

js/preload/preloadMain.js
KiranNiranjan: No, I don't think there is any other way and this is being used elsewhere

PR #373, opened 2018-05-21
ELECTRON-453: add functionality to ignore cert transparency checks

VishwasShashidhar:
## Description
Some customers would like to ignore certificate transparency checks for specific domains. This PR implements the required functionality [ELECTRON-453](https://perzoinc.atlassian.net/browse/ELECTRON-453)

## Approach
How does this change address the problem?
- #### Problem with the code: Currently, there's no way to ignore certificate transparency checks
- #### Fix: Add a new listener for ignoring certificate transparency checks on a browser window. Compare this with the whitelist configured and take necessary action.

## Learning
N/A

#### Blog Posts
N/A

## Related PRs
N/A

## Open Questions if any and Todos
- [x] Unit-Tests
[ELECTRON-453 Unit Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2021511/ELECTRON-453.Unit.Tests.pdf)
- [x] Documentation
- [x] Automation-Tests
[ELECTRON-453 Spectron Tests.pdf](https://github.com/symphonyoss/SymphonyElectron/files/2021512/ELECTRON-453.Spectron.Tests.pdf)

PR #553, opened 2019-01-22
ELECTRON-965 (Display warning dialog if certificate fails to verify Root CA or is not whitelisted)

KiranNiranjan:
## Description
Display warning dialog if the certificate fails to verify Root CA or is not whitelisted [ELECTRON-965](https://perzoinc.atlassian.net/browse/ELECTRON-965)

## Solution Approach
The certificate verification takes place in the following order

1. Root CA verifies and sends and `errorCode` of 0 for success
2. `ctWhiteList` is validated against hostName
3. Displays a warning dialog with the following message `Certificate Error: Cannot verify Root CA for the hostname: google.com`
 for the user to take it forward

### Screen shot
screenshot 2019-01-22 at 5 13 05 pm

## Related PRs
List related PRs against other branches / repositories:

branch | PR
------ | ------
SymphonyElectron (previous implementation) | [#373](https://github.com/symphonyoss/SymphonyElectron/pull/373)

## QA Checklist
- [X] Unit-Tests
screenshot 2019-01-22 at 5 10 56 pm

- [ ] Automation-Tests


js/windowMgr.js
```diff
@@ -70,6 +70,7 @@ const DEFAULT_HEIGHT = 600;
 
 // Certificate transparency whitelist
 let ctWhitelist = [];
+let ignoreAllCertErrors = false;
```
VishwasShashidhar: I think we should reset this to `true` if the browser window is refreshed because that's the behaviour on Chrome as well.

PR #576, opened 2019-02-28
protocol handle unit test

vrbsm:
## Description
protocol handle unit test

## QA Checklist
- [x] Unit-Tests
- [ ] Automation-Tests

### DIFF
--- a/spec/protocolHandler.spec.ts
+++ b/spec/protocolHandler.spec.ts
@@ -1,0 +1,75 @@
+jest.mock('../src/app/window-actions', () => {
+    return {
+        activate: jest.fn(),
+    };
+});
+
+jest.mock('../src/common/utils', () => {
+    return {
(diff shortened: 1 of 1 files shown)


spec/protocolHandler.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        isWindowsOS: false,
+    };
+});
+
+describe('protocol handler', () => {
+    let protocolHandlerInstance;
+
+    beforeEach(() => {
+        jest.resetModules();
+        const { protocolHandler } = require('../src/app/protocol-handler');
```
VishwasShashidhar: can we not use `import` here?

vrbsm: I am using require as a 'trick' to test private functions in typescript

PR #580, opened 2019-03-06
fix app path on windows

vrbsm:
## Description
fixing config handle unit test

## QA Checklist
- [x] Unit-Tests
- [ ] Automation-Tests

### DIFF
--- a/spec/__mocks__/electron.ts
+++ b/spec/__mocks__/electron.ts
@@ -1,5 +1,6 @@
 import { EventEmitter } from 'events';
 import * as path from 'path';
+import { isWindowsOS } from '../../src/common/env';
 const ipcEmitter = new EventEmitter();
 
 const mockIdleTime: number = 15;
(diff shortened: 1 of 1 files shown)


spec/__mocks__/electron.ts
```diff
@@ -28,17 +29,24 @@ interface IPowerMonitor {
     querySystemIdleTime(): void;
 }
 
-// use config provided by test framework
 const pathToConfigDir = (): string => {
     return path.join(__dirname, '/..') as string;
 };
 
+const pathToConfigDirWindows = (): string => {
```
VishwasShashidhar: do we need another method for this? Why not just have a condition in the main method to check if it's `Windows`

spec/__mocks__/electron.ts
```diff
@@ -28,17 +29,24 @@ interface IPowerMonitor {
...
+    return path.join(__dirname, '/../..') as string;
+};
+
 // electron app mock...
 const app: IApp = {
     getAppPath: pathToConfigDir,
     getPath: (type) => {
         if (type === 'exe') {
-            return path.join(pathToConfigDir(), executableName);
+            if (isWindowsOS) {
```
VishwasShashidhar: we can move this condition check to the `pathToConfigDir` method itself.

PR #603, opened 2019-03-19
utils unit test

vrbsm:
## Description
utils unit test
Screen Shot 2019-03-19 at 11 49 42

## QA Checklist
- [x] Unit-Tests
- [ ] Automation-Tests

### DIFF
--- a/spec/utils.spec.ts
+++ b/spec/utils.spec.ts
@@ -1,0 +1,122 @@
+import { compareVersions, formatString, getCommandLineArgs, getGuid, throttle } from '../src/common/utils';
+
+describe('utils', () => {
+    describe('`compareVersions`', () => {
+        it('should return -1 when v1 and v2 are invalid values', () => {
+            expect(compareVersions('1.0-1-1-', '2.1-1-1-')).toBe(-1);
+        });
+
+        it('should return -1 when v1 < v2', () => {
+            expect(compareVersions('v1.0', 'v2.1')).toBe(-1);
+        });
+
+        it('should return 1 when v1 > v2', () => {
+            expect(compareVersions('v1.8', 'v1.1')).toBe(1);
+        });
+
+        it('should return 0 when v1 is equal to v2', () => {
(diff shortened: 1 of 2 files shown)


src/common/utils.ts
```diff
@@ -182,7 +182,7 @@ export const throttle = (func: (...args) => void, wait: number): (...args) => vo
  * @param data {Object} - Data to be added
  *
  * @example
- * StringFormat(this will log {time}`, { time: '1234' })
+ * StringFormat('this will log {time}', [{ time: '1234' }])
```
KiranNiranjan: @vrbsm The 2nd arg should be `object`. As you can see we are using `for..in` loop

spec/utils.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            const expectedValue = `get-command-line-args: TypeError invalid func arg, must be an array: ${fakeValue}`;
+            expect(() => {
+                getCommandLineArgs(fakeValue as string[], argName, false);
+            }).toThrow(expectedValue);
+        });
+    });
+
+    describe('`formatString', () => {
+        const str = 'this will log {time}';
+        const data = [{time: '1234'}];
```
KiranNiranjan: This should be of type `object` and not `array`

spec/utils.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            expect(formatString(str)).toEqual(str);
+        });
+
+        it('should replace key to dynamics values when `formatString` is used', () => {
+            const expectedValue = 'this will log 1234';
+            expect(formatString(str, data)).toBe(expectedValue);
+        });
+
+        it('should return `str` when `data` not match', () => {
+            const dataTest = [{test: 123}];
```
KiranNiranjan: same here

src/common/utils.ts
KiranNiranjan: We need to fix the `formatString` method 🙂

PR #609, opened 2019-03-22
Typescript (Complete notification & notification settings modules)

KiranNiranjan:
## Description
Complete notification & notification settings modules

Screenshot 2019-03-22 at 2 25 36 PM

## QA Checklist
- [ ] Unit-Tests
- [ ] Automation-Tests


demo/index.html
```diff
@@ -281,7 +316,11 @@ <h1>Symphony Electron API Demo</h1>
             document.location.reload();
         });
 
-        window.addEventListener('message', (event) => console.log(event));
+        window.addEventListener('message', (event) => console.log(event.data));
```
VishwasShashidhar: will this print the notification text?

src/app/window-handler.ts
```diff
@@ -75,6 +77,25 @@ export class WindowHandler {
...
+        return {
+            width: 460,
+            height: 360,
+            show: false,
+            modal: true,
+            autoHideMenuBar: true,
+            webPreferences: {
+                sandbox: true,
+                nodeIntegration: false,
+                devTools: false,
```
VishwasShashidhar: should we also have `contextIsolation` as true?

src/app/window-handler.ts
```diff
@@ -258,7 +280,7 @@ export class WindowHandler {
                 return this.destroyAllWindow();
             }
 
-            if (this.config.minimizeOnClose) {
+            if (config.getConfigFields([ 'minimizeOnClose' ]).minimizeOnClose) {
```
VishwasShashidhar: why do we need this change?

src/renderer/notification-handler.ts
```diff
@@ -62,14 +65,15 @@ export default class NotificationHandler {
      */
     public setupNotificationPosition() {
         // This feature only applies to windows
-        if (isMac) {
+        if (!isMac || !app.isReady()) {
```
VishwasShashidhar: on `windows`, this condition will always be true and it'll be returned. Is this correct logic?

src/app/window-handler.ts
KiranNiranjan: `config.getConfigFields` always returns the latest from `config` file whereas `this.config` is only for initial read

src/renderer/notification-handler.ts
KiranNiranjan: fixed 😁

src/app/window-handler.ts
KiranNiranjan: added

PR #626, opened 2019-04-03
Typescript (Fix spellchecker issue and bump electron version)

KiranNiranjan:
## Description
Fix issues and add support for BC

### Changes
- Fixed spellchecker module
- Bumped up electron version to 5.x
- Update unit test case
- Updated app menu to support 5.x
- Fixed protocol handler

### Added
- Demo content for language changes
- Demo content for protocol handler

## QA Checklist
- [X] Unit-Tests
- [ ] Automation-Tests

### DIFF
--- a/src/app/crypto-handler.ts
+++ b/src/app/crypto-handler.ts
@@ -1,7 +1,7 @@
 import { app } from 'electron';
-import { Library } from 'ffi';
+import { Library } from 'ffi-napi';
 import * as path from 'path';
-import { refType, types } from 'ref';
+import { refType, types } from 'ref-napi';
 
 const execPath = path.dirname(app.getPath('exe'));
 import { isDevEnv, isMac } from '../common/env';
--- a/src/app/protocol-handler.ts
+++ b/src/app/protocol-handler.ts
@@ -1,5 +1,5 @@
 import { apiName } from '../common/api-interface';
-import { isMac, isWindowsOS } from '../common/env';
+import { isMac } from '../common/env';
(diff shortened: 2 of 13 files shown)


src/app/crypto-handler.ts
```diff
@@ -1,7 +1,7 @@
 import { app } from 'electron';
```
VishwasShashidhar: we don't need native encryption anymore. So, we might as well remove this.

src/app/protocol-handler.ts
```diff
@@ -61,7 +61,7 @@ class ProtocolHandler {
      */
     public processArgv(argv?: string[]): void {
         const protocolUriFromArgv = getCommandLineArgs(argv || process.argv, protocol.SymphonyProtocol, false);
-        if (isWindowsOS && protocolUriFromArgv) {
+        if (protocolUriFromArgv) {
```
VishwasShashidhar: reason for this change please?

VishwasShashidhar:
@KiranNiranjan Please rebase with master as well.

src/app/crypto-handler.ts
KiranNiranjan: Native encryption work if we disable `contextIsolation`

src/app/protocol-handler.ts
KiranNiranjan: We need to handle `protocol handler` from the cmd line for both `Windows` & `Mac` so removed this check

KiranNiranjan:
@VishwasShashidhar It's up to date with the master

PR #629, opened 2019-04-08
Typescript (Optimize loading styles & register global shortcuts)

KiranNiranjan:
## Description
Fix issues and add support for BC

### Added
- Implemented memory refresh logic

### Changes
- Optimized loading styles
- Register global shortcuts for `devtools`

## QA Checklist
- [X] Unit-Tests
- [ ] Automation-Tests

### DIFF
--- a/src/app/memory-monitor.ts
+++ b/src/app/memory-monitor.ts
@@ -46,38 +46,41 @@ class MemoryMonitor {
      * Validates the predefined conditions and refreshes the client
      */
     private validateMemory(): void {
+        const { memoryRefresh } = config.getConfigFields([ 'memoryRefresh' ]);
+        if (!memoryRefresh) {
+            logger.info(`memory refresh is disabled`);
+            return;
+        }
+
         (electron.powerMonitor as any).querySystemIdleTime((time) => {
             const idleTime = time * 1000;
-            if (!this.isInMeeting
+            if (!(!this.isInMeeting
                 && windowHandler.isOnline
                 && this.canReload
                 && idleTime > this.maxIdleTime
-                && (this.memoryInfo && this.memoryInfo.private > this.memoryThreshold)
+                && (this.memoryInfo && this.memoryInfo.private > this.memoryThreshold))
             ) {
-                const { memoryRefresh } = config.getConfigFields([ 'memoryRefresh' ]);
-                if (memoryRefresh) {
-                    const mainWindow = windowHandler.getMainWindow();
-                    if (mainWindow && windowExists(mainWindow)) {
-                        logger.info(`Reloading the app to optimize memory usage as
-                            memory consumption is ${this.memoryInfo.private}kb is greater than? ${this.memoryThreshold}kb threshold
-                            system idle tick was ${idleTime}ms is greater than ${this.maxIdleTime}ms
-                            user was in a meeting? ${this.isInMeeting}
-                            is network online? ${windowHandler.isOnline}`);
-                        windowHandler.setIsAutoReload(true);
-                        mainWindow.reload();
-                        this.canReload = false;
-                        setTimeout(() => {
-                            this.canReload = true;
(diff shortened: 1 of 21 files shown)


src/app/window-handler.ts
```diff
@@ -315,11 +334,18 @@ export class WindowHandler {
             }
         });
 
+        this.mainWindow.once('closed', () => {
+            this.destroyAllWindow();
```
VishwasShashidhar: please refactor the method name to be `destroyAllWindows` in place of `destroyAllWindow`

src/app/window-handler.ts
```diff
@@ -760,6 +786,43 @@ export class WindowHandler {
...
+
+    /**
+     * Verifies and toggle devtool based on global config settings
+     * else displays a dialog
+     */
+    private onRegisterDevtools(): void {
+        const focusedWindow = BrowserWindow.getFocusedWindow();
+        const { devToolsEnabled } = config.getGlobalConfigFields([ 'devToolsEnabled' ]);
+        if (focusedWindow && windowExists(focusedWindow)) {
+            if (devToolsEnabled) {
```
VishwasShashidhar: we can move this condition to the above if block

KiranNiranjan: @VishwasShashidhar we cannot because first, we need to validate if the window exists and then validate if devTools enabled

vrbsm:
@KiranNiranjan travis failed. take a look

VishwasShashidhar:
@KiranNiranjan Tests related to activity detection are failing, please fix them.

src/app/memory-monitor.ts
```diff
@@ -0,0 +1,88 @@
...
+    private validateMemory(): void {
+        (electron.powerMonitor as any).querySystemIdleTime((time) => {
+            const idleTime = time * 1000;
+            if (!this.isInMeeting
+                && windowHandler.isOnline
+                && this.canReload
+                && idleTime > this.maxIdleTime
+                && (this.memoryInfo && this.memoryInfo.private > this.memoryThreshold)
+            ) {
+                const { memoryRefresh } = config.getConfigFields([ 'memoryRefresh' ]);
```
VishwasShashidhar: we can move this logic before the `if` condition and exit if memory refresh is turned off.

src/app/memory-monitor.ts
```diff
@@ -0,0 +1,88 @@
...
+                if (memoryRefresh) {
```
VishwasShashidhar: same as above, please move this check (inverted) before the above `if` condition.

src/app/memory-monitor.ts
```diff
@@ -0,0 +1,88 @@
...
+                            is network online? ${windowHandler.isOnline}`);
+                        windowHandler.setIsAutoReload(true);
+                        mainWindow.reload();
+                        this.canReload = false;
+                        setTimeout(() => {
+                            this.canReload = true;
+                        }, this.memoryRefreshThreshold);
+                    }
+                }
+            } else {
```
VishwasShashidhar: please avoid `if else` statements and use early returns.

src/app/window-handler.ts
VishwasShashidhar: let's avoid `if else` chain and use early return logic

src/renderer/app-bridge.ts
```diff
@@ -68,7 +68,12 @@ export default class AppBridge {
         const { method, data } = event.data;
         switch (method) {
             case apiCmds.getVersionInfo:
-                this.broadcastMessage('get-version-info-callback', ssf.getVersionInfo());
+                ssf.getVersionInfo().then((versionInfo) => {
```
VishwasShashidhar: can we not use `async await` here?

KiranNiranjan:
New commit: Fixed unit tests. Thanks @vrbsm

PR #649, opened 2019-05-16
ELECTRON-1247 (Remove old Symphony launch agent plist file)

KiranNiranjan:
## Description
Remove old Symphony launch agent plist file 
[ELECTRON-1247](https://perzoinc.atlassian.net/browse/ELECTRON-1247)

## Solution Approach
Using the third party `auto-launch` module (used previously) to remove the plist file.
The new launch on startup API provided by Electron creates an entry @ login items

### DIFF
--- a/js/autoLaunch/index.js
+++ b/js/autoLaunch/index.js
@@ -19,6 +19,7 @@ const props = isMac ? {
     path: getAutoLaunchPath() || process.execPath,
 };
 
+// TODO: Remove this method in the future
 removeOldLaunchAgent();
 
 /**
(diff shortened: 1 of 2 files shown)


js/autoLaunch/index.js
```diff
@@ -18,6 +19,17 @@ const props = isMac ? {
     path: getAutoLaunchPath() || process.execPath,
 };
 
+/**
```
VishwasShashidhar: Please add this to a function and call the function, it'll be cleaner.

PR #654, opened 2019-05-21
ELECTRON-1173: open browser tab for same host url

VishwasShashidhar:
## Description
Currently, we are opening the same host url from message list in a new window instead of the default browser tab. This PR fixes that issue.

## Solution Approach
Compare the new window url with that of the main url (pod url) and check if the new window url starts with the pod url, if so, open the url in a browser.

## Related PRs
N/A

### DIFF
--- a/src/app/child-window-handler.ts
+++ b/src/app/child-window-handler.ts
@@ -9,7 +9,7 @@ import { getGuid } from '../common/utils';
 import { config } from './config-handler';
 import { monitorWindowActions, removeWindowEventListener } from './window-actions';
 import { ICustomBrowserWindow, windowHandler } from './window-handler';
-import { getBounds, handleCertificateProxyVerification, injectStyles, preventWindowNavigation, } from './window-utils';
+import { getBounds, handleCertificateProxyVerification, injectStyles, preventWindowNavigation } from './window-utils';
 
 const DEFAULT_POP_OUT_WIDTH = 300;
 const DEFAULT_POP_OUT_HEIGHT = 600;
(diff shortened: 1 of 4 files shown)


src/app/child-window-handler.ts
```diff
@@ -9,12 +9,7 @@ import { getGuid } from '../common/utils';
 import { config } from './config-handler';
 import { monitorWindowActions, removeWindowEventListener } from './window-actions';
 import { ICustomBrowserWindow, windowHandler } from './window-handler';
-import {
-    getBounds,
-    handleCertificateProxyVerification,
-    injectStyles,
-    preventWindowNavigation,
-} from './window-utils';
+import { getBounds, handleCertificateProxyVerification, injectStyles, preventWindowNavigation } from './window-utils';
```
KiranNiranjan: This change is not required

src/app/window-handler.ts
```diff
@@ -16,16 +15,8 @@ import { handleChildWindow } from './child-window-handler';
...
-    getBounds,
-    handleCertificateProxyVerification,
-    handleDownloadManager,
-    injectStyles,
-    isSymphonyReachable,
-    preventWindowNavigation,
-    windowExists,
-} from './window-utils';
+import { createComponentWindow, getBounds, handleCertificateProxyVerification, handleDownloadManager, injectStyles, isSymphonyReachable, preventWindowNavigation, windowExists } from './window-utils';
+import DesktopCapturerSource = Electron.DesktopCapturerSource;
```
KiranNiranjan: This change is not required

src/app/child-window-handler.ts
VishwasShashidhar: I think it's because of the formatting. Do we have a formatting configuration for `TS`?

PR #655, opened 2019-05-21
Typescript (Fix custom user data path and menu items)

KiranNiranjan:
## Description
Fix custom user data path and menu item arrangements

## Solution Approach
- Config: Create a init file to initialize `userData` path before the app start 
- Menu: Rearrange and include memory refresh menu item

## Screenshot
Screenshot 2019-05-21 at 3 30 47 PM

### DIFF
--- a/src/app/init.ts
+++ b/src/app/init.ts
@@ -1,0 +1,22 @@
+import { app } from 'electron';
+import * as path from 'path';
+
+import { isDevEnv } from '../common/env';
+import { getCommandLineArgs } from '../common/utils';
+
+// Handle custom user data path from process.argv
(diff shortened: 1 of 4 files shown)


src/app/init.ts
```diff
@@ -0,0 +1,22 @@
+import { app } from 'electron';
+import * as path from 'path';
+
+import { isDevEnv } from '../common/env';
+import { getCommandLineArgs } from '../common/utils';
+
+// Handle custom user data path from process.argv
+const userDataPathArg: string | null = getCommandLineArgs(process.argv, '--userDataPath=', false);
```
VishwasShashidhar: Can we not just have this file's entire logic in a method in `main.ts`

KiranNiranjan: No, because there are other `import` statements that get initiated

PR #659, opened 2019-05-22
Set env path on app start child_process issue

keerthi16:
## Description
Set env path on app start child_process issue refer -> `https://github.com/electron/electron/issues/7688`

## Solution Approach
N/A

## Related PRs
N/A

### DIFF
--- a/src/app/init.ts
+++ b/src/app/init.ts
@@ -1,7 +1,8 @@
 import { app } from 'electron';
 import * as path from 'path';
+import * as shellPath from 'shell-path';
 
-import { isDevEnv } from '../common/env';
+import { isDevEnv, isMac } from '../common/env';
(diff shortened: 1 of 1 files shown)


src/app/init.ts
```diff
@@ -18,5 +19,26 @@ if (userDataPath) {
...
 }
 
+// Setting the env path child_process issue https://github.com/electron/electron/issues/7688
+(async () => {
+    try {
+        const paths = await shellPath();
+        if (paths) {
+            return process.env.PATH = paths;
+        }
+        process.env.PATH = [
```
VishwasShashidhar: is this applicable only for macOS? If so, let's add the check

keerthi16: Done 👍

PR #679, opened 2019-06-11
feat: ELECTRON-1297 (Add role alert to the notification container)

KiranNiranjan:
## Description
Add role alert to the notification container to make screenreader read it properly 
[ELECTRON-1297](https://perzoinc.atlassian.net/browse/ELECTRON-1297)

## Solution Approach
Add `role: 'alert'` to the notification container


femontanha:
Nice one! 👍

PR #681, opened 2019-06-13
revert: ffi and version changes

keerthi16:
## Description
Contest-isolation was not merged to R55.2 so reverting this changes

## Solution Approach
N/A

## Related PRs
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,12 +83,12 @@
   },
   "devDependencies": {
     "@types/enzyme": "3.9.0",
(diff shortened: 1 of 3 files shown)


keerthi16:
@VishwasShashidhar If we merge the context-isolation PR on the SFE in R55.2 we will not be needing this PR.
And if that is not the case please merge the PR and will be changing this back to 1.55.x.

PR #686, opened 2019-06-17
fix: ELECTRON-1238 (Handle session permissions)

KiranNiranjan:
## Description
Handle session permissions
[ELECTRON-1238](https://perzoinc.atlassian.net/browse/ELECTRON-1238)

## Solution Approach
Verify global config's permissions field and prevent access to some properties

Notes: Add support for handling session permissions for media and others

### DIFF
--- a/src/app/window-actions.ts
+++ b/src/app/window-actions.ts
@@ -237,10 +237,10 @@ export const handleSessionPermissions = (permission: boolean, message: string, c
  */
 export const handlePermissionRequests = (webContents: Electron.webContents): void => {
 
-    const { session } = webContents;
-    if (!webContents || !session) {
+    if (!webContents || !webContents.session) {
         return;
     }
+    const { session } = webContents;
 
     const { permissions } = config.getGlobalConfigFields([ 'permissions' ]);
     if (!permissions) {
--- a/src/app/window-actions.ts
+++ b/src/app/window-actions.ts
@@ -1,11 +1,24 @@
-import { BrowserWindow } from 'electron';
+import { BrowserWindow, dialog } from 'electron';
+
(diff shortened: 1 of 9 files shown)


src/app/window-actions.ts
```diff
@@ -194,3 +207,65 @@ export const removeWindowEventListener = (window: BrowserWindow): void => {
...
+
+/**
+ * Sets permission requests for the window
+ *
+ * @param webContents {Electron.webContents}
+ */
+export const handlePermissionRequests = (webContents: Electron.webContents): void => {
+
+    const { session } = webContents;
+    if (!webContents || !session) {
```
keerthi16: `webContents` needs to be checked first

src/app/window-actions.ts
```diff
@@ -194,3 +207,65 @@ export const removeWindowEventListener = (window: BrowserWindow): void => {
...
+ *
+ * @param permission {boolean} - config value to a specific permission
+ * @param message {string} - custom message displayed to the user
+ * @param callback {function}
+ */
+export const handleSessionPermissions = (permission: boolean, message: string, callback: (permission: boolean) => void): void => {
+    logger.info(`window-action: permission is ->`, { type: message, permission });
+
+    if (!permission) {
+        const fullMessage = `${i18n.t('Your administrator has disabled')()} ${message}. ${i18n.t('Please contact your admin for help')()}`;
```
VishwasShashidhar: we'll need a space after `disabled`

KiranNiranjan: @VishwasShashidhar There is already space after `disabled`

VishwasShashidhar: Ah! Templating, yes, thanks :)

PR #689, opened 2019-06-19
fix: ELECTRON-1264 (Fix ATS for localhost and update TLS version to 1.2)

KiranNiranjan:
## Description
Fix ATS for localhost and update TLS version to `1.2`
[ELECTRON-1264](https://perzoinc.atlassian.net/browse/ELECTRON-1264)

## Solution Approach
- Disable sending crash dumps to server

To prevent vulnerable in the adjacent network
ATS for the localhost has been updated with 
- NSTemporaryExceptionAllowsInsecureHTTPLoads - NO
- NSTemporaryExceptionMinimumTLSVersion - 1.2

### Updated Plist file
```
<key>NSAppTransportSecurity</key>
<dict>
   <key>NSExceptionDomains</key>
   <dict>
      <key>localhost</key>
      <dict>
         <key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
         <false />
         <key>NSTemporaryExceptionMinimumTLSVersion</key>
         <string>1.2</string>
      </dict>
   </dict>
   <key>NSAllowsLocalNetworking</key> // note: can't be updated via electron-builder
   <true />
   <key>NSAllowsArbitraryLoads</key> // note: can't be updated via electron-builder
   <true />
</dict>
```

### Learning
https://github.com/electron-userland/electron-builder/issues/3377

Notes: Fix ATS for localhost and update TLS version to 1.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,17 @@
     "appId": "com.symphony.electron-desktop",
     "mac": {
       "target": "dmg",
-      "category": "public.app-category.business"
+      "category": "public.app-category.business",
+      "extendInfo": {
+        "NSAppTransportSecurity": {
+          "NSExceptionDomains": {
+            "localhost": {
+              "NSTemporaryExceptionAllowsInsecureHTTPLoads": false,
+              "NSTemporaryExceptionMinimumTLSVersion": "1.2"
+            }
+          }
+        }
+      }
     },
     "dmg": {
       "contents": [
--- a/package.json
+++ b/package.json
@@ -49,17 +49,7 @@
     "appId": "com.symphony.electron-desktop",
     "mac": {
       "target": "dmg",
-      "category": "public.app-category.business",
-      "extendInfo": {
-        "NSAppTransportSecurity": {
-          "NSExceptionDomains": {
-            "localhost": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -49,7 +49,17 @@
     "appId": "com.symphony.electron-desktop",
     "mac": {
       "target": "dmg",
-      "category": "public.app-category.business"
+      "category": "public.app-category.business",
+      "extendInfo": {
```
VishwasShashidhar: I don't think we'll need this since we set the minimum version via the Chrome flags. Let's remove this for now and add it later only if required.

src/app/window-handler.ts
```diff
@@ -190,7 +190,8 @@ export class WindowHandler {
 
         try {
             const extra = { podUrl: this.globalConfig.url, process: 'main' };
-            crashReporter.start({ ...this.globalConfig.crashReporter, extra });
+            const defaultOpts = { uploadToServer: false, companyName: 'Symphony', submitURL: '' };
```
VishwasShashidhar: Have you tested that the crash reports still get generated?

KiranNiranjan: @VishwasShashidhar Yes, Tested it and the crash report will generate

PR #691, opened 2019-06-24
feat: ELECTRON-1332 (Make Context Isolation configurable)

KiranNiranjan:
## Description
Make Context Isolation configurable
[ELECTRON-1332](https://perzoinc.atlassian.net/browse/ELECTRON-1332)

## Solution Approach
Add a flag `contextIsolation: false` in `Symphony.config` global configuration file and use that to set the context isolation feature on all the browser windows.

Notes: Make Context Isolation configurable via global config file

### DIFF
--- a/config/Symphony.config
+++ b/config/Symphony.config
@@ -8,6 +8,7 @@
     "isCustomTitleBar": true,
     "memoryRefresh": true,
     "devToolsEnabled": true,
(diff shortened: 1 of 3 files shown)


VishwasShashidhar:
@KiranNiranjan Let's abstract the `opts` and customise as required for the specific windows.

PR #696, opened 2019-07-02
fix: ELECTRON-1243: fix chrome flags tests

VishwasShashidhar:
## Description
Add support for disabling throttling and background rendering.

[ELECTRON-1243](https://perzoinc.atlassian.net/browse/ELECTRON-1243)

## Solution
Add a flag `disableThrottling` in `Symphony.config` that would allow users to disable throttling.

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -69,6 +69,7 @@ export class WindowHandler {
     public spellchecker: SpellChecker | undefined;
 
     private readonly contextIsolation: boolean;
+    private readonly backgroundThrottling: boolean;
     private readonly windowOpts: ICustomBrowserWindowConstructorOpts;
     private readonly globalConfig: IConfig;
     private readonly config: IConfig;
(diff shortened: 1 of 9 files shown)


src/app/window-handler.ts
```diff
@@ -89,9 +90,12 @@ export class WindowHandler {
         // Use these variables only on initial setup
         this.config = config.getConfigFields([ 'isCustomTitleBar', 'mainWinPos', 'minimizeOnClose', 'notificationSettings' ]);
         this.globalConfig = config.getGlobalConfigFields([ 'url', 'contextIsolation' ]);
+        const { customFlags } = config.getGlobalConfigFields([ 'customFlags' ]) as IConfig;
```
KiranNiranjan: This can be merged with the above line

src/app/window-handler.ts
```diff
@@ -89,9 +90,12 @@ export class WindowHandler {
...
+        this.globalConfig.customFlags = customFlags;
 
         this.windows = {};
         this.contextIsolation = this.globalConfig.contextIsolation || false;
+        this.backgroundThrottling = !this.globalConfig.customFlags.disableThrottling;
```
KiranNiranjan: we can read `disableThrottling` directly from `customFlags`

src/app/window-handler.ts
VishwasShashidhar: Addressed

src/app/window-handler.ts
VishwasShashidhar: Addressed as part of the above comment.

PR #703, opened 2019-07-06
feat: ELECTRON-1357: show version info from pod

VishwasShashidhar:
## Description
To keep the versioning consistent across all products in Symphony, we need to show the main pod version in the **About Symphony** window on windows & macOS. This change implements that logic.
[ELECTRON-1357](https://perzoinc.atlassian.net/browse/ELECTRON-1357)

Screenshot 2019-07-06 at 18 34 42

Screenshot 2019-07-06 at 18 36 56

## Solution Approach
N/A

## Related PRs
N/A


src/app/window-handler.ts
```diff
@@ -405,7 +406,15 @@ export class WindowHandler {
             if (!this.aboutAppWindow || !windowExists(this.aboutAppWindow)) {
                 return;
             }
-            this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion, version });
+            let { clientVersion, buildNumber }: IVersionInfo = versionHandler.getVersionInfo();
+            this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion });
+            eventEmitter.on('update-version-info', (versionInfo) => {
```
KiranNiranjan: I think instead of `Event Emitters`, we can fetch the data if calls `getVersionInfo()` has missing values `clientVersion` & `buildNumber`. This also prevents multiple calls to get the client info

`eventEmitter` here can cause a memory leak if the about window is open and closed multiple times

src/app/window-handler.ts
```diff
@@ -426,7 +435,13 @@ export class WindowHandler {
             if (!this.moreInfoWindow || !windowExists(this.moreInfoWindow)) {
                 return;
             }
-            this.moreInfoWindow.webContents.send('more-info-data');
+            const verInfo: IVersionInfo = versionHandler.getVersionInfo();
+            this.moreInfoWindow.webContents.send('more-info-data', verInfo);
+            eventEmitter.on('update-version-info', (versionInfo) => {
```

src/app/version-handler.ts
```diff
@@ -0,0 +1,105 @@
...
+        get(url, (res) => {
+
+            let body: string = '';
+            res.on('data', (d: Buffer) => {
+                body += d;
+            });
+
+            res.on('end', () => {
+                try {
+                    const data = JSON.parse(body)[0];
```
KiranNiranjan: We can cache the results here to prevent multiple calls

src/app/main.ts
```diff
@@ -22,11 +23,25 @@ let isAppAlreadyOpen: boolean = false;
...
+        copyright,
+    });
+};
+
 if (isMac) {
-    // Sets application version info that will be displayed in about app panel
-    app.setAboutPanelOptions({ applicationVersion: `${clientVersion}-${version}`, version: buildNumber });
+    let { clientVersion, buildNumber }: IVersionInfo = versionHandler.getVersionInfo();
+    setAboutPanel(clientVersion, buildNumber);
+    eventEmitter.on('update-version-info', (versionInfo) => {
```
KiranNiranjan: same as below

src/app/version-handler.ts
```diff
@@ -0,0 +1,105 @@
+import { get } from 'https';
```
KiranNiranjan: I think we can use the `net` module which has better support for proxy

src/app/window-handler.ts
VishwasShashidhar: there are just 2 listeners attached to the event emitter. not sure how that can cause memory leak. can you elaborate?

and, the idea to use event emitter is because we are sending an asynchronous request to get the data, so, we'll need to update the version data once when we initialise the app and after we get the version info from the server.

src/app/version-handler.ts
VishwasShashidhar: You mean this module `https://nodejs.org/docs/latest-v10.x/api/net.html#net_net`?

KiranNiranjan: No, [this](https://github.com/electron/electron/blob/master/docs/api/net.md) module

src/app/window-handler.ts
KiranNiranjan: `eventEmitter.on` keeps adding listeners whenever the window is opened.

### With EventEmitter
```let { clientVersion, buildNumber }: IVersionInfo = versionHandler.getVersionInfo();
this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion });
eventEmitter.on('update-version-info', (versionInfo) => {
	clientVersion = versionInfo.clientVersion;
	buildNumber = versionInfo.buildNumber;
	if (this.aboutAppWindow) {
	    this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion });
	}
});
```

### This same as above without EventEmitter
```let { clientVersion, buildNumber }: IVersionInfo = versionHandler.getVersionInfo();
this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion });
if (!clientVersion || !buildNumber) {
	const { clientVersion, buildNumber }: = versionHandler.getClientVersion();
	this.aboutAppWindow.webContents.send('about-app-data', { buildNumber, clientVersion });
}
```

src/app/main.ts
```diff
@@ -34,24 +36,13 @@ const setAboutPanel = (clientVersion: string, buildNumber: string) => {
...
-
-// Electron sets the default protocol
-app.setAsDefaultProtocolClient('symphony');
-
 /**
  * Main function that init the application
  */
 const startApplication = async () => {
     await app.whenReady();
+    const versionInfo: IVersionInfo = await versionHandler.getClientVersion();
```
KiranNiranjan: This will prevent the application from launching if there any network related issues

VishwasShashidhar: Thanks Kiran for the observation. I've corrected it in the next commit.

src/app/version-handler.ts
```diff
@@ -0,0 +1,115 @@
...
+
+            const hostname = nodeURL.parse(podUrl).hostname;
+            const protocol = nodeURL.parse(podUrl).protocol;
+            const versionApiPath = '/webcontroller/HealthCheck/version/advanced';
+
+            const url = `${protocol}//${hostname}${versionApiPath}`;
+            logger.info(`Trying to get version info for the URL: ${url}`);
+
+            const request = net.request(url);
+            request.on('response', (res) => {
```
keerthi16: @VishwasShashidhar We should also listen for `request.on('error')` else this would display an error dialog.
Screenshot 2019-07-08 at 12 45 48 PM

VishwasShashidhar: Good point @keerthi16, I've addressed the above.

PR #707, opened 2019-07-09
fix: ELECTRON-1338: fix download manager issues

VishwasShashidhar:
## Description
We had the following problems with the download manager
- We used to get a pop-up dialog every time we download a file. This was a regression for not using `electron-dl`
- If the same file was downloaded twice, the download manager was not updating the filename count appropriately
[ELECTRON-1338](https://perzoinc.atlassian.net/browse/ELECTRON-1338)

## Solution Approach
- Use `electron-dl` that automatically downloads the file to the user's download folder on the local system
- Update count in react state

## Related PRs
N/A

### DIFF
--- a/src/app/main.ts
+++ b/src/app/main.ts
@@ -1,5 +1,6 @@
 import { app } from 'electron';
 
+import * as electronDownloader from 'electron-dl';
 import { buildNumber, clientVersion, version } from '../../package.json';
 import { isDevEnv, isMac } from '../common/env';
 import { logger } from '../common/logger';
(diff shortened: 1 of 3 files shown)


src/app/main.ts
```diff
@@ -1,6 +1,7 @@
 import { app } from 'electron';
 import * as shellPath from 'shell-path';
 
+import * as electronDownloader from 'electron-dl';
```
KiranNiranjan: This should be grouped with the above import statements

VishwasShashidhar: Good catch 👍 Fixed it.

PR #708, opened 2019-07-09
fix: ELECTRON-1037 (Disable context menu for notification windows)

KiranNiranjan:
## Description
Disable context menu for notification windows
[ELECTRON-1037](https://perzoinc.atlassian.net/browse/ELECTRON-1037)

## Solution Approach
Add context menu listener for notification DOM element and prevent default


VishwasShashidhar:
@KiranNiranjan Can we simply not use this? https://electronjs.org/docs/api/menu#render-process

KiranNiranjan:
@VishwasShashidhar Yes, That's how it is currently being handled. I'm attaching the listeners for the notification content instead of the `window` object

PR #717, opened 2019-07-11
fix: ELECTRON-1346 & ELECTRON-1370 (Fix notification text & background color)

KiranNiranjan:
## Description
Fixes notification text & background color
[ELECTRON-1346](https://perzoinc.atlassian.net/browse/ELECTRON-1346)
[ELECTRON-1370](https://perzoinc.atlassian.net/browse/ELECTRON-1370)

## Solution Approach
Convert colors to `hex` which fixes the issue

## Before
Screenshot 2019-07-11 at 3 22 52 PM

## After
Screenshot 2019-07-11 at 3 20 45 PM

### DIFF
--- a/src/renderer/components/notification-comp.tsx
+++ b/src/renderer/components/notification-comp.tsx
@@ -53,8 +53,8 @@ export default class NotificationComp extends React.Component<{}, IState> {
      */
     public render(): JSX.Element {
         const { title, company, body, image, icon, id, color } = this.state;
-        const colorHex = color ? '#' + color : undefined;
-        const isLightTheme = colorHex ? colorHex.match(whiteColorRegExp) : true;
+        const colorHex = (color && !color.startsWith('#')) ? '#' + color : color;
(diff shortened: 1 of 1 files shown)


src/renderer/components/notification-comp.tsx
```diff
@@ -53,8 +53,8 @@ export default class NotificationComp extends React.Component<{}, IState> {
      */
     public render(): JSX.Element {
         const { title, company, body, image, icon, id, color } = this.state;
-        const colorHex = color ? '#' + color : undefined;
-        const isLightTheme = colorHex ? colorHex.match(whiteColorRegExp) : true;
+        const colorHex = (color && !color.startsWith('#')) ? '#' + color : color;
```
VishwasShashidhar: Is this check because we don't get `#` with colour codes from SFE at times?

KiranNiranjan: Yes, that's correct

PR #727, opened 2019-07-12
fix: ELECTRON-1382 (Fix notification z-index when alwaysOnTop is enabled)

KiranNiranjan:
## Description
Fixes notification z-index when alwaysOnTop is enabled
[ELECTRON-1382](https://perzoinc.atlassian.net/browse/ELECTRON-1382)

## Solution Approach
Bring active notifications to front

Notes: Fixes notification z-index when alwaysOnTop is enabled

### DIFF
--- a/src/renderer/notification.ts
+++ b/src/renderer/notification.ts
@@ -301,6 +301,21 @@ class Notification extends NotificationHandler {
     }
 
     /**
+     * Brings all the notification to the top
+     * issue: ELECTRON-1382
+     */
+    public moveNotificationToTop(): void {
+        const notificationWindows = this.activeNotifications as ICustomBrowserWindow[];
+        notificationWindows
+            .filter((browserWindow) => typeof browserWindow.notificationData === 'object' && browserWindow.isVisible())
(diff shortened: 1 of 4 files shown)


src/renderer/notification.ts
```diff
@@ -300,6 +300,21 @@ class Notification extends NotificationHandler {
...
+    /**
+     * Brings all the notification to the top
+     * issue: ELECTRON-1382
+     */
+    public moveNotificationToTop(): void {
+        const notificationWindows = this.activeNotifications as ICustomBrowserWindow[];
+        notificationWindows
+            .filter((browserWindow) => typeof browserWindow.notificationData === 'object' && browserWindow.isVisible())
+            .forEach((browserWindow) => {
+                if (browserWindow && windowExists(browserWindow) && browserWindow.isVisible()) {
```
keerthi16: browserWindow.isVisible() not required here.

KiranNiranjan: `isVisible` is required here because the notification window can be hidden any movement while in the loop

PR #736, opened 2019-07-17
fix: ELECTRON-1406: fix issue in windows menu popping up for custom title bar

VishwasShashidhar:
## Description
Fix issue in windows menu popping up for custom title bar
[ELECTRON-1406](https://perzoinc.atlassian.net/browse/ELECTRON-1406)

## Solution Approach
Add an extra check (to see if we have custom title bar) while triggering the pop up menu when "Alt" key is pressed.

### DIFF
--- a/src/app/window-actions.ts
+++ b/src/app/window-actions.ts
@@ -20,7 +20,6 @@ enum Permissions {
     OPEN_EXTERNAL = 'openExternal',
 }
 const PERMISSIONS_NAMESPACE = 'Permissions';
-const { isCustomTitleBar } = config.getUserConfigFields([ 'isCustomTitleBar' ]);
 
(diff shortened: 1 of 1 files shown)


src/app/window-actions.ts
```diff
@@ -20,6 +20,7 @@ enum Permissions {
     OPEN_EXTERNAL = 'openExternal',
 }
 const PERMISSIONS_NAMESPACE = 'Permissions';
+const { isCustomTitleBar } = config.getUserConfigFields([ 'isCustomTitleBar' ]);
```
KiranNiranjan: `isCustomTitleBar` is a public instance and can be accessed directly from `windowHandler.isCustomTitleBar`

VishwasShashidhar: Noted, will change 👍

PR #743, opened 2019-07-19
fix: ELECTRON-1431: add logic to update pod version dynamically

VishwasShashidhar:
## Description
When we redirect to a different POD, currently, the pod version doesn't get updated dynamically. This PR fixes that issue.
[ELECTRON-1431](https://perzoinc.atlassian.net/browse/ELECTRON-1431)

## Solution Approach
When a user logs in to a different POD, we register our electron specific APIs on SFE, we use this workflow to also update the pod version data on SDA.

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -16,7 +16,7 @@ import { config, IConfig } from './config-handler';
 import { SpellChecker } from './spell-check-handler';
 import { checkIfBuildExpired } from './ttl-handler';
 import DesktopCapturerSource = Electron.DesktopCapturerSource;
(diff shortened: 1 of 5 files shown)


src/app/window-handler.ts
```diff
@@ -640,8 +661,8 @@ export class WindowHandler {
                 resizable: false,
                 alwaysOnTop: true,
             }, {
-                devTools: false,
-            }), ...{ winKey: streamId },
+                    devTools: false,
```
KiranNiranjan: This indentation makes it difficult to read.

PR #752, opened 2019-07-27
fix: ELECTRON-1451 (Fix notification flashing issue and logo in network error)

KiranNiranjan:
## Description
Fix notification flashing issue and logo in network error
[ELECTRON-1451](https://perzoinc.atlassian.net/browse/ELECTRON-1451)

## Solution Approach
- Reset the notification flash interval
- update the new Symphony logo

#### Network Error
Screenshot 2019-07-27 at 9 30 45 AM

### DIFF
--- a/src/renderer/components/notification-comp.tsx
+++ b/src/renderer/components/notification-comp.tsx
@@ -160,6 +160,9 @@ export default class NotificationComp extends React.Component<{}, IState> {
         const { color, flash } = data;
         data.color = (color && !color.startsWith('#')) ? '#' + color : color;
         this.setState(data as IState);
+        if (this.flashTimer) {
+            clearInterval(this.flashTimer);
+        }
         if (flash) {
             const origColor = data.color;
             this.flashTimer = setInterval(() => {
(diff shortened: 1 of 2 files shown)


src/renderer/components/notification-comp.tsx
```diff
@@ -160,6 +160,9 @@ export default class NotificationComp extends React.Component<{}, IState> {
         const { color, flash } = data;
         data.color = (color && !color.startsWith('#')) ? '#' + color : color;
         this.setState(data as IState);
+        if (this.flashTimer) {
```
keerthi16: @KiranNiranjan This is just an improvement. If there is a `this.flashTimer` do not run this part [line](https://github.com/symphonyoss/SymphonyElectron/pull/752/files#diff-5e55b4298403f384ef25fde3a18882caR166)
So this will prevent from flashing again if we push a new notification.

KiranNiranjan: @keerthi16 That's correct but what if there is the same tag with a different color? So, it will always be better to reset.

PR #762, opened 2019-08-07
fix: ELECTRON-1465 (Persist locale changes to the user config file)

KiranNiranjan:
## Description
Persist locale changes to the user config file
[ELECTRON-1465](https://perzoinc.atlassian.net/browse/ELECTRON-1465)

## Solution Approach
- Save the locale changes to the user config file
- Use the locale from the config file for the initial launch

### DIFF
--- a/src/common/i18n.ts
+++ b/src/common/i18n.ts
@@ -52,6 +52,9 @@ class Translation {
      * @param locale {LocaleType}
      */
     public isValidLocale(locale: LocaleType): boolean {
+        if (!locale) {
+            return false;
+        }
         const localeMatch: string[] | null = locale.match(localeCodeRegex);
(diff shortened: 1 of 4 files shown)


src/common/i18n.ts
```diff
@@ -47,6 +46,16 @@ class Translation {
...
     }
 
+    /**
+     * Validates the locale using Regex
+     *
+     * @param locale {LocaleType}
+     */
+    public isValidLocale(locale: LocaleType): boolean {
+        const localeMatch: string[] | null = locale.match(localeCodeRegex);
+        return !(!locale && (!localeMatch || localeMatch.length < 1));
```
keerthi16: `!locale` check should be made first. `locale.match` will fail if undefined.

PR #775, opened 2019-08-30
fix: ELECTRON-1505 (Prevent opening links longer than 2083)

KiranNiranjan:
## Description
Prevent opening links longer than 2083
[ELECTRON-1505](https://perzoinc.atlassian.net/browse/ELECTRON-1505)

## Solution Approach
- `PreventDefault` when the URL length is longer than 2083

## Related PRs

branch | PR
------ | ------
SymphonyElectron | [#774](https://github.com/symphonyoss/SymphonyElectron/pull/774)

### DIFF
--- a/src/app/child-window-handler.ts
+++ b/src/app/child-window-handler.ts
@@ -195,6 +195,7 @@ export const handleChildWindow = (webContents: WebContents): void => {
         } else {
             if (newWinUrl && newWinUrl.length > 2083) {
                 logger.info(`child-window-handler: new window url length is greater than 2083, not performing any action!`);
+                event.preventDefault();
                 return;
             }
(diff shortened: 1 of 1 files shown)


src/app/child-window-handler.ts
```diff
@@ -195,6 +195,7 @@ export const handleChildWindow = (webContents: WebContents): void => {
         } else {
             if (newWinUrl && newWinUrl.length > 2083) {
                 logger.info(`child-window-handler: new window url length is greater than 2083, not performing any action!`);
+                event.preventDefault();
```
VishwasShashidhar: we can refactor this! Just move `event.preventDefault()` to the top of the `else` block and remove the redundant one.

PR #807, opened 2019-11-14
fix: SDA-1588 (Add support for chromium sandbox mode)

KiranNiranjan:
## Description
Add support for the chromium sandbox mode
[SDA-1588](https://perzoinc.atlassian.net/browse/SDA-1588)

## Solution Approach
- Enable chromium sandbox
- Add sandbox support for all the renderer process
- This terminates a stale process running when there is a termination of the main process

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "main": "lib/src/app/init.js",
   "types": "lib/src/app/init.d.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


KiranNiranjan:
@johankwarnmarksymphony **Mattias** I will merge this for the time being if you guys have any concerns we can discuss and revert this PR. Thanks

PR #809, opened 2019-11-15
feat: SDA-945 Logs: Unify logs collection procedure

mattias-symphony:
## Description
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SDA-945)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/src/app/reports-handler.ts
+++ b/src/app/reports-handler.ts
@@ -4,7 +4,7 @@ import * as electron from 'electron';
 import * as fs from 'fs';
 import * as path from 'path';
 
-import { ILogFile } from '../common/api-interface';
+import { ILogFile, IRetrieveLogs } from '../common/api-interface';
 import { isLinux, isMac } from '../common/env';
 import { i18n } from '../common/i18n';
 import { logger } from '../common/logger';
@@ -108,13 +108,9 @@ export const exportLogsFinalize = ( logs: ILogFile[] ): void => {
         });
 };
 
-class RetrieveLogs {
-    public webContents: Electron.WebContents | undefined = undefined;
-}
-
-const retrieveLogs = new RetrieveLogs();
-
-export { retrieveLogs };
+export const retrieveLogs: IRetrieveLogs = {
+    webContents: undefined,
+};
 
 /**
  * Compress and export logs stored under system log directory
--- a/src/app/reports-handler.ts
+++ b/src/app/reports-handler.ts
@@ -4,6 +4,7 @@ import * as electron from 'electron';
 import * as fs from 'fs';
 import * as path from 'path';
 
+import { ILogFile } from '../common/api-interface';
 import { isLinux, isMac } from '../common/env';
 import { i18n } from '../common/i18n';
 import { logger } from '../common/logger';
@@ -63,7 +64,7 @@ const generateArchiveForDirectory = (source: string, destination: string, fileEx
  * MacOS - /Library/Logs/Symphony/
  * Windows - AppData\Roaming\Symphony\logs
  */
-export const exportLogs = (): void => {
+export const exportLogsFinalize = ( logs: ILogFile[] ): void => {
     const FILE_EXTENSIONS = [ '.log' ];
     const MAC_LOGS_PATH = '/Library/Logs/Symphony/';
     const LINUX_LOGS_PATH = '/.config/Symphony/';
@@ -82,6 +83,11 @@ export const exportLogs = (): void => {
         });
         return;
     }
+
+    for (const log of logs) {
+        fs.writeFileSync(source + ( !isMac && !isLinux ? '\\' : '' ) + log.filename + '.log', log.contents );
+    }
+
     const destPath = (isMac || isLinux) ? '/logs_symphony_' : '\\logs_symphony_';
     const timestamp = new Date().getTime();
     const destination = app.getPath('downloads') + destPath + timestamp + '.zip';
@@ -100,6 +106,28 @@ export const exportLogs = (): void => {
                 });
             }
         });
+};
+
+class RetrieveLogs {
+    public webContents: Electron.WebContents | undefined = undefined;
+}
+
+const retrieveLogs = new RetrieveLogs();
+
+export { retrieveLogs };
+
+/**
+ * Compress and export logs stored under system log directory
+ *
+ * MacOS - /Library/Logs/Symphony/
+ * Windows - AppData\Roaming\Symphony\logs
+ */
+export const exportLogs = (): void => {
+    if (typeof retrieveLogs.webContents === 'undefined') {
+        exportLogsFinalize([]);
+    } else {
+        retrieveLogs.webContents.send('retrieve-logs');
+    }
 };
 
 /**
--- a/src/app/reports-handler.ts
+++ b/src/app/reports-handler.ts
@@ -85,7 +85,7 @@ export const exportLogsFinalize = ( logs: ILogFile[] ): void => {
     }
 
     for (const log of logs) {
-        fs.writeFileSync(source + ( !isMac && !isLinux ? '\\' : '' ) + log.filename + '.log', log.contents );
+        fs.writeFileSync(path.join( source, log.filename + '.log' ), log.contents );
     }
 
     const destPath = (isMac || isLinux) ? '/logs_symphony_' : '\\logs_symphony_';
--- a/src/renderer/app-bridge.ts
+++ b/src/renderer/app-bridge.ts
@@ -56,6 +56,7 @@ export class AppBridge {
         ): void => this.gotMediaSource(error, source),
         onNotificationCallback: (event, data) => this.notificationCallback(event, data),
         onAnalyticsEventCallback: (data) => this.analyticsEventCallback(data),
+        onRegisterLogRetrieverCallback: () => this.logRetrieverCallback(),
     };
 
     constructor() {
@@ -85,6 +86,7 @@ export class AppBridge {
         switch (method) {
             case apiCmds.getVersionInfo:
                 const versionInfo = await ssf.getVersionInfo();
+                alert( JSON.stringify( versionInfo ) );
                 this.broadcastMessage('get-version-info-callback', {
                     requestId: data.requestId,
                     response: versionInfo,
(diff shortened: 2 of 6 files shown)


src/renderer/app-bridge.ts
```diff
@@ -85,6 +86,7 @@ export class AppBridge {
         switch (method) {
             case apiCmds.getVersionInfo:
                 const versionInfo = await ssf.getVersionInfo();
+                alert( JSON.stringify( versionInfo ) );
```
keerthi16: Why do we need this?

mattias-symphony: No, that was a mistake, I meant to remove it :) It is gone now.

src/app/reports-handler.ts
```diff
@@ -82,6 +83,11 @@ export const exportLogs = (): void => {
         });
         return;
     }
+
+    for (const log of logs) {
+        fs.writeFileSync(source + ( !isMac && !isLinux ? '\\' : '' ) + log.filename + '.log', log.contents );
```
keerthi16: We can use `path.join` instead of this ` !isMac && !isLinux ? '\\' : '' `

mattias-symphony: Good idea, I will change it to that :)

src/app/reports-handler.ts
```diff
@@ -102,6 +108,28 @@ export const exportLogs = (): void => {
         });
 };
 
+class RetrieveLogs {
```
keerthi16: Do we really need to have a class?? we can just export the variable.

mattias-symphony: I tried, but I couldn't get it to work, I kept getting errors. Probably me doing something wrong... What would that look like?

keerthi16: in `src/app/reports-handler.ts`
```
export const retrieveLogs: IRetrieveLogs = {
    webContents: undefined
};
```

Create `IRetrieveLogs` in `src/common/api-interface.ts`
```
export interface IRetrieveLogs {
    webContents: Electron.WebContents | undefined,
}
```

src/app/reports-handler.ts
```diff
@@ -82,6 +83,11 @@ export const exportLogs = (): void => {
         });
         return;
     }
+
+    for (const log of logs) {
+        fs.writeFileSync(path.join( source, log.filename + '.log' ), log.contents );
```
johanblumenberg-symphony: `ILogFile.filename` should contain the log file extension as well

src/renderer/ssf-api.ts
```diff
@@ -236,6 +237,28 @@ export class SSFApi {
         }
     }
 
+    /**
+     * Allows JS to register a log retriever that can be used by electron main process.
+     * @param  {Object} retriever  function that can be called to retrieve logs
+     */
+    public registerLogRetriever(retriever): void {
```
johanblumenberg-symphony: Put a type on the argument

src/renderer/ssf-api.ts
```diff
@@ -236,6 +237,28 @@ export class SSFApi {
...
+        if (typeof retriever === 'function') {
+            local.logRetriever = retriever;
```
johanblumenberg-symphony: What if you register two log retrievers?

src/renderer/ssf-api.ts
```diff
@@ -236,6 +237,28 @@ export class SSFApi {
...
+            local.ipcRenderer.send(apiName.symphonyApi, {
+                cmd: apiCmds.registerLogRetriever,
+            });
+        }
+    }
+
+    /**
+     * Allows JS to send logs to the electron main process when it calls the log retriever.
+     */
+    public logReceiver(logs): void {
```
johanblumenberg-symphony: type on the argument

src/renderer/ssf-api.ts
```diff
@@ -236,6 +237,28 @@ export class SSFApi {
         }
     }
 
+    /**
+     * Allows JS to register a log retriever that can be used by electron main process.
+     * @param  {Object} retriever  function that can be called to retrieve logs
```
johanblumenberg-symphony: No need to put the type in the jsdoc comment, since it is typescript the type can be taken from the method signature

src/renderer/ssf-api.ts
johanblumenberg-symphony: What is the contract for the retriever function? Does it have to call `logRecevier`? How many times?

If so, I think it would be better to have the function return the logs, maybe as a promise to allow for asynchronous collection of logs. The interface will be much easier to understand, and it's harder to forget to return logs by mistake.

src/renderer/app-bridge.ts
```diff
@@ -153,7 +154,13 @@ export class AppBridge {
                     ssInstance.handleMessageEvents(data);
                 }
                 break;
-        }
+            case apiCmds.registerLogRetriever:
+                ssf.registerLogRetriever(this.callbackHandlers.onRegisterLogRetrieverCallback);
+                break;
+            case apiCmds.logReceiver:
+                ssf.logReceiver( data.logs );
```
KiranNiranjan: Extra blank space is not required for arguments

src/renderer/ssf-api.ts
```diff
@@ -236,6 +237,28 @@ export class SSFApi {
...
+            });
+        }
+    }
+
+    /**
+     * Allows JS to send logs to the electron main process when it calls the log retriever.
+     */
+    public logReceiver(logs): void {
+        local.ipcRenderer.send(apiName.symphonyApi, {
+            cmd: apiCmds.logReceiver, logs,
```
KiranNiranjan: please push `logs` into a new line

src/app/reports-handler.ts
mattias-symphony: The existing code which packages log files only includes files with .log extension. So if we want to send the extension with the filename, should we then extract the extension of log files and add to the list of log files to be packaged? It is not clear to me if this might have other implications?

src/renderer/ssf-api.ts
mattias-symphony: I don't think there is a use-case for that though. But ok, what change is needed here?

src/renderer/ssf-api.ts
mattias-symphony: When the log retriever is called, the logs are sent all in one go as a message. Not sure I follow how you want it to be done instead - do you know of any place in the code where it's done like that which I could look at for reference?

src/app/reports-handler.ts
johanblumenberg-symphony: I  might want to send .json files, .har, or whatever that might be useful logs

src/renderer/ssf-api.ts
johanblumenberg-symphony: Yes there is. We want to use this interface from RTC. If some other app later want to add some logs, for example SFE-Lite or the email extension, that will not work.

If we can only have a single log receiver, we need to document that properly, and we need to write the code to collect logs from multiple sources somewhere else. But I think this would be the best place to do that.

src/app/reports-handler.ts
mattias-symphony: But I assume there is a reason the "generateArchiveForDirectory" was restricted to only include .log files? If we allow for any extensions, it means any files that happen to be in the source folder will be included (not just those we send with retrieve logs), instead of just the .log files as we do today. Which might be what we want, I am just pointing out that it is a change in behaviour that I have no idea what impact it could have.  Also, you could still send a .json file, it will just get an extra ".log" extension when saved.

src/renderer/ssf-api.ts
mattias-symphony: When sending the logs, it is a list of log files, so we could send multiple logs (gathered from multiple sources if needed) whenever the retriever is invoked.

src/app/reports-handler.ts
mattias-symphony: ok, yes, that works, thanks :)

src/renderer/ssf-api.ts
mattias-symphony: done

src/renderer/ssf-api.ts
mattias-symphony: done

src/renderer/app-bridge.ts
mattias-symphony: done

src/renderer/ssf-api.ts
mattias-symphony: done

src/renderer/ssf-api.ts
mattias-symphony: What type should it have? I did it the same way as the existing registerProtocolHandler which looks like this:
`public registerProtocolHandler(protocolHandler): void`

src/app/reports-handler.ts
johanblumenberg-symphony: Can't you just send any file in that folder that ends wit h`.log`, **and** any file returned in the new API?

src/renderer/ssf-api.ts
johanblumenberg-symphony: Yes, but there would have to be some code in, for example, SFE-Lite that would gather the log files from different sources in the app. And the same code would have to be put in SFE-Client-App.

It would be better to have this API support multiple users within the app. That's usually how you do browser APIs.

src/renderer/ssf-api.ts
johanblumenberg-symphony: Instead of registering one callback of type `void retreiveLogs()` that is required to call another global method, `sendLogs()`, have the callback function return the logs.

That reduces the number of global methods in the API, and makes the callback easier to get right, since it is clear from the signature that it should return the logs.

It also avoids the problem of how to handle calls to `sendLogs()` when no one is collecting any logs. Because there won't be any method.
It will also help to correlate the number of calls to registered log sources, and their responses.

src/app/reports-handler.ts
mattias-symphony: No, the API for creating the zip file works on a whole folder.

src/renderer/ssf-api.ts
mattias-symphony: Ok, so what would you suggest we do? How should the API look?

src/renderer/ssf-api.ts
mattias-symphony: Ok, that make sense, but I don't know what that would look like.

PR #817, opened 2019-11-27
feat: Implemented unified log extraction

mattias-symphony:
## Description

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()


src/demo/index.html
```diff
@@ -437,6 +473,15 @@ <h1>Symphony Electron API Demo</h1>
             }
         });
 
+        /**
+         * Log retriever
+         */
+        if (window.ssf) {
```
johanblumenberg-symphony: Do we need to implement both ways of sending logs? Both the ssf javascript method, and the postMessage() method? I thong we were not to use the old one any more, because it doesn't work with different contexts. Is it not enough to implement the postMessage() way? Will anyone use the other one?

src/demo/index.html
```diff
@@ -408,9 +420,33 @@ <h1>Symphony Electron API Demo</h1>
...
+
         window.addEventListener('message', (event) => {
             const { data, method } = event.data;
             switch (method) {
+                case 'collect-logs':
+                    if( data === 'RTC' ) {
+                        if (window.ssf) {
+                            window.ssf.logReceiver('RTC', getRTCLogs());
+                        } else {
+                            postMessage( apiCmds.logReceiver, { logName: 'RTC', logFiles: getRTCLogs() } );
```
johanblumenberg-symphony: I think it is less confusing as an example to just hardcode the string 'log-receiver' here, instead of having to look up a constant somewhere. It's only going to be used in one place anyway.

src/demo/index.html
```diff
@@ -155,6 +165,8 @@ <h1>Symphony Electron API Demo</h1>
                 activate: 'activate',
                 registerBoundsChange: 'register-bounds-change',
                 registerProtocolHandler: 'register-protocol-handler',
+                registerLogRetriever: 'register-log-retriever',
+                logReceiver: 'log-receiver',
```
johanblumenberg-symphony: Shouldn't the operation be a verb? I think `send-logs` would be better.

src/demo/index.html
```diff
@@ -105,6 +105,16 @@ <h1>Symphony Electron API Demo</h1>
         <p>snippet output:</p>
         <image id='snippet-img'/>
 
+        <hr>
+        <p>RTC Logs:</p>
```
johanblumenberg-symphony: I think you should remove RTC from the example. Both from the text and from the element IDs. This feature is unaware who is adding logs. Name it something generic instead.

src/demo/index.html
```diff
@@ -408,9 +420,33 @@ <h1>Symphony Electron API Demo</h1>
...
+                return [ { filename: filename1, contents: contents1 }, 
+                    { filename: filename2, contents: contents2 }, ];
+            }
+        }
+
         window.addEventListener('message', (event) => {
             const { data, method } = event.data;
             switch (method) {
+                case 'collect-logs':
+                    if( data === 'RTC' ) {
```
johanblumenberg-symphony: Are you broadcasting one message for each registered provider? Wouldn't it be easier if each provider responds to the same message? It's a broadcast, so it will reach all.

I don't think there is a case where you wouldn't want everyone to provide logs, you would always send to everyone anyway.
The implementation is simpler, because you don't have to check that the message is for you.

src/app/reports-handler.ts
```diff
@@ -102,6 +142,36 @@ export const exportLogs = (): void => {
         });
 };
 
+export const finalizeLogExports = (logs: ILogs) => {
+    receivedLogs.push(logs);
+
+    let allReceived = true;
+    for (const logRetriever of logRetrievers ) {
+        let found = false;
```
johanblumenberg-symphony: `let found = receivedLogs.some(log => log.logName === logRetreiver.logName)`

src/renderer/ssf-api.ts
```diff
@@ -263,6 +264,33 @@ export class SSFApi {
...
     }
 
+    /**
+     * Allows JS to register a log retriever that can be used by the
+     * electron main process to retrieve current logs.
+     */
+    public registerLogRetriever(collectLogs: (logName: string) => void, logName: string): void {
+        if (typeof collectLogs === 'function') {
+
+            local.collectLogsCallback = collectLogs;
```
johanblumenberg-symphony: Will this be overwritten if there are multiple log providers?

src/demo/index.html
mattias-symphony: I'm not sure, my understanding was we needed both for compatibility. Maybe @KiranNiranjan would have more details.

src/demo/index.html
mattias-symphony: Yeah, I'll change it to just the string

src/demo/index.html
mattias-symphony: Yes, I'll go with send-logs

src/demo/index.html
mattias-symphony: Will do

src/demo/index.html
mattias-symphony: Not really sure what this means, I was pretty much just following existing patterns. What change do you propose?

src/app/reports-handler.ts
mattias-symphony: Cool :)

src/renderer/ssf-api.ts
mattias-symphony: Ah, right, this should probably be a list?

src/demo/index.html
KiranNiranjan: Yes, we definitely need to support both as currently, 

1. We have this feature configurable so, this should work if users turn off context isolation.
2. There will be a new API that we are going to be using in the latest versions of Electron `6.x`+ that requires the support for SSF APIs

src/app/main-api-handler.ts
```diff
@@ -53,6 +54,12 @@ ipcMain.on(apiName.symphonyApi, (event: Electron.IpcMainEvent, arg: IApiArgs) =>
             // we make use of it and update the pod version info on SDA
             windowHandler.updateVersionInfo();
             break;
+        case apiCmds.registerLogRetriever:
+            registerLogRetriever( event.sender, arg.logName );
+            break;
+        case apiCmds.logReceiver:
+            finalizeLogExports( arg.logs );
```
KiranNiranjan: Please remove extra spaces for arguments

src/app/reports-handler.ts
```diff
@@ -17,19 +18,30 @@ import { logger } from '../common/logger';
...
+                if (fs.existsSync( file )) {
+                    fs.unlinkSync(file);
+                }
+            }
             logger.info(`reports-handler: generated archive for directory ${source}`);
             return resolve();
         });
 
         archive.on('error', (err: Error) => {
+            for (const file of filesForCleanup) {
```
KiranNiranjan: If for some reason archive fails won't we lose the actual logfile?

src/app/reports-handler.ts
```diff
@@ -53,17 +65,45 @@ const generateArchiveForDirectory = (source: string, destination: string, fileEx
...
+const receivedLogs: ILogs[] = [];
+
+export const registerLogRetriever = ( sender: Electron.WebContents, logName: string): void => {
+    logRetrievers.push( { sender, logName } );
+};
+
+export const collectLogs = (): void => {
+    receivedLogs.length = 0;
+    for (const logRetriever of logRetrievers ) {
+        logRetriever.sender.send('collect-logs', logRetriever.logName );
```
KiranNiranjan: Remove extra space

src/app/reports-handler.ts
```diff
@@ -102,6 +142,36 @@ export const exportLogs = (): void => {
         });
 };
 
+export const finalizeLogExports = (logs: ILogs) => {
+    receivedLogs.push(logs);
+
+    let allReceived = true;
+    for (const logRetriever of logRetrievers ) {
```
KiranNiranjan:
```suggestion
    for (const logRetriever of logRetrievers) {
```

src/app/reports-handler.ts
```diff
@@ -53,17 +65,45 @@ const generateArchiveForDirectory = (source: string, destination: string, fileEx
...
 
+interface ILogRetriever {
+    sender: Electron.WebContents;
+    logName: string;
+}
+
+const logRetrievers: ILogRetriever[] = [];
+const receivedLogs: ILogs[] = [];
+
+export const registerLogRetriever = ( sender: Electron.WebContents, logName: string): void => {
```
KiranNiranjan:
```suggestion
export const registerLogRetriever = (sender: Electron.WebContents, logName: string): void => {
```

src/app/reports-handler.ts
mattias-symphony: Yes, I think this is desired behavior? These log files are just kept in memory and we only want to include them in the archive, not have them stored in file.

KiranNiranjan: No, these log files are stored in userData path and if the archive fails we lose both the file and archive right?

mattias-symphony: No, these logs were received in response to 'collect-logs' being sent. They are not stored anywhere, just kept in memory. The only reason I save them as temp files is to be able to add them to the archive - after archiving is done, the temp files needs to be removed whether we succeeded or not.

It would be different if we had opted for a general "receive logs as we go and store them with the electron logs", but that isn't what we want. Instead, when the user goes to the menu and selects "Show logs in..." at that point we must retrieve the current in-memory logs, and add them to the same archive as we add the electron logs to.

src/demo/index.html
johanblumenberg-symphony: Now you pass `postMessage('collect-logs') once for each listener, and all listeners will receive that message. Each listener is expected to only reply to one of them.

I propose that you call `postMessage('collect-logs')` only once. All listeners will receive that message, and they can respond with `send-logs`.

src/app/main-api-handler.ts
VishwasShashidhar: +1

src/app/reports-handler.ts
```diff
@@ -17,19 +18,30 @@ import { logger } from '../common/logger';
...
 
     return new Promise((resolve, reject) => {
         logger.info(`reports-handler: generating archive for directory ${source}`);
         const output = fs.createWriteStream(destination);
         const archive = archiver('zip', { zlib: { level: 9 } });
+        const filesForCleanup: string[] = [];
 
         output.on('close', () => {
+            for (const file of filesForCleanup) {
+                if (fs.existsSync( file )) {
```
VishwasShashidhar: Please fix the formatting here.

src/demo/index.html
VishwasShashidhar: If we decide to change the name in the future, it'd be hard to refactor. I think it'll be better if we have the string in a variable itself.

johanblumenberg-symphony: @VishwasShashidhar This is a demo, which other people are supposed to look at to know how to use the API, right? So it should be as simple as possible. It should only include exactly what is needed to demonstrate the feature. No unnecessary details.

VishwasShashidhar: While I agree in this use case @johankwarnmarksymphony, we should still from a refactoring POV, have a variable. In any case, like you said, this is the demo page, so, we can proceed with the current implementation.

PR #819, opened 2019-11-29
feat: SDA-1592 screen share indicator frame

mattias-symphony:
## Description
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SDA-1592)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()


finos-admin:
Thank you for your pull request and welcome to our community. We require contributors to be identified and we couldn't parse GitHub user details of the following contributors: **Mattias Gustavsson**.

In order for your PR to be reviewed and merged, please follow the following directions:
1. check if your git client is configured with an email to sign commits `git config --list | grep email`
2. If not, set it up using `git config --global user.email email@example.com`
3. Make sure that the git commit email is configured in your GitHub account settings, see https://github.com/settings/emails
4. Amend the authors in the commit history; use `git commit --amend` to only change last commit, or one between `git reset --soft` and `git rebase` to checkout changes, rewrite commit history locally and (force) push changes to the downstream branch

Project team: please **do not merge this Pull Request** until Foundation staff have confirmed that a CLA is in place for the new contributor(s) listed above.

If you still have any questions or just don't know what to do next, please email [help@finos.org](mailto:help@finos.org). /CC @finos-admin

src/app/window-handler.ts
```diff
@@ -4,6 +4,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { format, parse } from 'url';
 
+import { ChildProcess, ExecException, execFile } from 'child_process';
```
KiranNiranjan: I think it is better to move this `screenShareIndicatorFrame` module to a separate file instead of handling everything in `window-handler.ts`

installer/win/Symphony-x64.aip
```diff
@@ -174,6 +175,7 @@
     <ROW File="Jobber.exe" Component_="Jobber.exe" FileName="Jobber.exe" Attributes="0" SourcePath="..\..\dist\win-unpacked\resources\app.asar.unpacked\node_modules\spawn-rx\vendor\jobber\Jobber.exe" SelfReg="false" NextFile="keyboardlayoutmanager.node" DigSign="true"/>
     <ROW File="LICENSE.electron.txt" Component_="blink_image_resources_200_percent.pak" FileName="LICENS~1.TXT|LICENSE.electron.txt" Attributes="0" SourcePath="..\..\dist\win-unpacked\LICENSE.electron.txt" SelfReg="false" NextFile="LICENSES.chromium.html"/>
     <ROW File="LICENSES.chromium.html" Component_="blink_image_resources_200_percent.pak" FileName="LICENS~1.HTM|LICENSES.chromium.html" Attributes="0" SourcePath="..\..\dist\win-unpacked\LICENSES.chromium.html" SelfReg="false" NextFile="natives_blob.bin"/>
+    <ROW File="ScreenShareIndicatorFrame.exe" Component_="ScreenShareIndicatorFrame.exe" FileName="SCREEN~2.EXE|ScreenShareIndicatorFrame.exe" Attributes="0" SourcePath="..\..\node_modules\screen-share-indicator-frame\ScreenShareIndicatorFrame.exe" SelfReg="false" DigSign="true"/>
```
KiranNiranjan: I think it is better to use the `dist` path for this file instead of from `node_module`

src/app/window-handler.ts
```diff
@@ -514,6 +525,16 @@ export class WindowHandler {
             this.addWindow(opts.winKey, this.screenPickerWindow);
         });
         ipcMain.once('screen-source-selected', (_event, source) => {
+            if (isWindowsOS) {
+                const str = JSON.stringify(source);
+                logger.info(`window-handler: screen-source-selected, source: ${str} id: ${id}!`);
```
KiranNiranjan: logger's internal method stringify objects when passed as an argument
```suggestion
                logger.info(`window-handler: screen-source-selected`, source, id);
```

src/app/window-handler.ts
```diff
@@ -709,11 +730,19 @@ export class WindowHandler {
...
             displays.forEach((element) => {
                 if (displayId === element.id.toString()) {
-                    this.createScrenSharingFrameWindow('screen-sharing-frame',
-                    element.workArea.width,
-                    element.workArea.height,
-                    element.workArea.x,
-                    element.workArea.y);
+                    if (isWindowsOS) {
+                        const str = JSON.stringify(element);
+                        logger.info(`window-handler: MG element: ${str}`);
```
```suggestion
                        logger.info(`window-handler: MG element:`, element);
```

installer/win/Symphony-x64.aip
mattias-symphony: I just copied how the ScreenSnippet works (next line) as the indicator frame is the same type of thing. Does this mean ScreenSnippet is wrong as well? Also, it seems to work fine - what is the advantage of "dist path"?

src/app/window-handler.ts
mattias-symphony: No, it is not much code, and moving it would make it more complicated, as it ties in with the window handling. I think it is better to have it as part of window-handler

src/app/window-handler.ts
mattias-symphony: ok, cool, I'll change it

src/app/window-handler.ts
VishwasShashidhar: @mattias-symphony Problem arrives when we want to write unit tests, that's the reason we try to componentize everything. My suggestion is same as @KiranNiranjan's as well.

But, let's also get @johankwarnmarksymphony's thoughts.

mattias-symphony: Then I would suggest that at the point when we actually write the unit test, that would be a good time to refactor into a separate module.

johankwarnmarksymphony: The screen share indicator frame is a separate windows program, so we already have it separated. I think we could wait with separating the java-script for now.

installer/win/Symphony-x64.aip
VishwasShashidhar: @mattias-symphony We can use either one of the paths, but `dist` represents the distributable folder, hence we choose that. You are right that ScreenSnippet.exe is also pointed to `node_modules`, we'll need to fix that. If you see the x86.aip file, we point everything to the `dist` folder, so, we'll benefit from being consistent in the two files. I suspect that ScreenSnippet.exe might be accidentally pointing to `node_modules` instead of `dist` in this case.

src/app/window-handler.ts
VishwasShashidhar: @johankwarnmarksymphony Can you please explain why we'd want to postpone something we'd eventually do? We might as well do it now if we plan to do it in the future. Just my 2 cents.

installer/win/Symphony-x64.aip
mattias-symphony: Ok, so I change both to use dist path

mattias-symphony: @keerthi16 @VishwasShashidhar Actually, this will not work - it looks like for the ScreenSnippet tool, it is not actually copied to the dist folder, it only lives in the node_modules, where it is build. So I guess that is why the installer points to node_modules for that one as well. So any suggestions how we would fix this?

VishwasShashidhar: @mattias-symphony It's available in `dist\win-unpacked\resources\app.asar.unpacked\node_modules\screen-snippet\bin\Release\ScreenSnippet.exe`

mattias-symphony: Ah, yes, it is there :) I'll change it to use that

installer/win/Symphony-x64.aip
```diff
@@ -175,7 +175,7 @@
...
     <ROW File="LICENSES.chromium.html" Component_="blink_image_resources_200_percent.pak" FileName="LICENS~1.HTM|LICENSES.chromium.html" Attributes="0" SourcePath="..\..\dist\win-unpacked\LICENSES.chromium.html" SelfReg="false" NextFile="natives_blob.bin"/>
-    <ROW File="ScreenShareIndicatorFrame.exe" Component_="ScreenShareIndicatorFrame.exe" FileName="SCREEN~2.EXE|ScreenShareIndicatorFrame.exe" Attributes="0" SourcePath="..\..\node_modules\screen-share-indicator-frame\ScreenShareIndicatorFrame.exe" SelfReg="false" DigSign="true"/>
+    <ROW File="ScreenShareIndicatorFrame.exe" Component_="ScreenShareIndicatorFrame.exe" FileName="SCREEN~2.EXE|ScreenShareIndicatorFrame.exe" Attributes="0" SourcePath="..\..\dist\win-unpacked\resources\app.asar.unpacked\node_modules\screen-share-indicator-frame\ScreenShareIndicatorFrame.exe" SelfReg="false" DigSign="true"/>
     <ROW File="ScreenSnippet.exe" Component_="ScreenSnippet.exe" FileName="SCREEN~1.EXE|ScreenSnippet.exe" Attributes="0" SourcePath="..\..\node_modules\screen-snippet\bin\Release\ScreenSnippet.exe" SelfReg="false" NextFile="indexvalidatorx64.exe" DigSign="true"/>
```
VishwasShashidhar: @mattias-symphony Can you update this too?

PR #830, opened 2019-12-12
chore: SDA-1623: switch NodeJs dynamically during build

VishwasShashidhar:
## Description
Add support to switch nodejs dynamically during the build
[SDA-1623](https://perzoinc.atlassian.net/browse/SDA-1623)

## Solution Approach
Use NVM to switch between Node versions

## Related PRs
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "productName": "Symphony",
   "version": "6.0.0",
   "clientVersion": "2.0.1",
-  "buildNumber":" 0",
+  "buildNumber": "0",
   "searchAPIVersion": "1.55.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "Symphony",
   "version": "6.0.0",
   "clientVersion": "2.0.1",
-  "buildNumber": "0",
+  "buildNumber":" 0",
```
KiranNiranjan: Is this change required?

VishwasShashidhar: Weird! I'll fix that.

PR #831, opened 2019-12-13
fix: New screen snippet tool

mattias-symphony:
## Description
[JIRA-ticket](https://perzoinc.atlassian.net/browse/SDA-1616)

## Solution Approach

## Related PRs

branch | PR
------ | ------
ScreenSnippet2 | [link](https://github.com/symphonyoss/ScreenSnippet2/pull/1)


src/renderer/ssf-api.ts
```diff
@@ -332,6 +332,17 @@ export class SSFApi {
         }
     }
 
+    /**
+     * Cancel a screen capture in progress
+     *
+     * @param screenSnippetCallback {function}
+     */
+    public closeScreenSnippet(): void {
```
KiranNiranjan: I don't think we need to expose this API. 
Closing the snippet should be handled on the SDA side and an exception to be thrown

mattias-symphony: This is in preparation for https://perzoinc.atlassian.net/browse/SDA-1284 which will require the frontend to issue a "closeScreenSnippet "command when the user leaves the "Create wall post" dialog.

KiranNiranjan: ok 👍

package.json
```diff
@@ -132,8 +132,8 @@
     "shell-path": "2.1.0"
   },
   "optionalDependencies": {
-    "screen-snippet": "git+https://github.com/symphonyoss/ScreenSnippet.git#v1.0.8",
-    "screen-share-indicator-frame": "https://github.com/symphonyoss/ScreenShareIndicatorFrame.git#v1.0.0",
+    "screen-snippet": "git+https://github.com/mattias-symphony/ScreenSnippet2.git",
```
VishwasShashidhar: Can we point to a specific version please? We can manage releases in ScreenSnippet2. I'd also prefer if the Project name was something like `WindowsSnippingTool` or along those lines as opposed to just `ScreenSnippet2`

src/app/screen-snippet-handler.ts
```diff
@@ -67,6 +67,22 @@ class ScreenSnippet {
...
+     * @param webContents {Electron.webContents}
+     */
+    public async cancelCapture() {
+        logger.info(`screen-snippet-handler: Cancel screen capture!`);
+        this.focusedWindow = BrowserWindow.getFocusedWindow();
+
+        try {
+            await this.execCmd(this.captureUtil, []);
+        } catch (error) {
+            logger.error(`screen-snippet-handler: screen capture cancedl failed with error: ${error}!`);
```
VishwasShashidhar: typo in `cancedl`

package.json
mattias-symphony: As soon as the C/C++ pull request is approved and I can merge it, I will give it version 1.0.0, make a release and point to the main repo instead of to my fork.

src/app/screen-snippet-handler.ts
mattias-symphony: Yes, I will fix :)

package.json
mattias-symphony: I am not sure we should change the name - the previous snippet tool was also windows only, so calling the new one ScreenSnippet2 keeps the naming consistent. Also, I am not sure what problems it would cause to change the name of a forked repo. Would we need to merge to master first? I just don't see much benefit, if any, in changing it at this point, and I can see there being a cost to it

VishwasShashidhar: My thought process is that we should have a proper naming convention :) Not sure why the old project was named `ScreenSnippet` as it's quite generic. I thought we can correct that here.

But, for now, we can keep this.

PR #844, opened 2020-01-08
fix: SDA-1657 Maximized SDA saves co-ordinates

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1657 SIG: Maximized SDA never saves the location


src/app/window-handler.ts
```diff
@@ -144,16 +144,20 @@ export class WindowHandler {
         this.spellchecker = new SpellChecker();
         logger.info(`window-handler: initialized spellchecker module with locale ${this.spellchecker.locale}`);
 
+        logger.info('window-handler: createApplication mainWinPos: ' + JSON.stringify(this.config.mainWinPos));
```
mattias-symphony: I think you don't need to  do JSON.stringify here? Just doing ${this.config.mainWinPos} should work?

VishwasShashidhar: @mattias-symphony `mainWinPos` is an object, so, stringily is required actually.

PR #857, opened 2020-01-23
fix: SDA-1708 Move screen-sharing indicator to the same screen as Symphony…

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1708 CLONE - RTC -  Screen sharing indicator always shown on primary screen when sharing Application on second screen

Move screen-sharing indicator to the same screen as Symphony Application

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -440,7 +440,7 @@ export class WindowHandler {
     /**
      * Move window to the same screen as main window
      */
-    public moveWindow(windowToMove: BrowserWindow) {
+    public moveWindow(windowToMove: BrowserWindow, fixedYPosition?: number) {
         if (this.mainWindow && windowExists(this.mainWindow)) {
             const display = electron.screen.getDisplayMatching(this.mainWindow.getBounds());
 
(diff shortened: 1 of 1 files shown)


src/app/window-handler.ts
```diff
@@ -785,6 +792,7 @@ export class WindowHandler {
         }
 
         this.screenSharingIndicatorWindow = createComponentWindow('screen-sharing-indicator', opts);
+        this.moveWindow(this.screenSharingIndicatorWindow, topPositionOfIndicatorScreen);
```
KiranNiranjan: @johankwarnmarksymphony `topPositionOfIndicatorScreen` where is this value coming from?

KiranNiranjan: Got it, It's a constant with a value of `16`

src/app/window-handler.ts
```diff
@@ -440,7 +440,7 @@ export class WindowHandler {
     /**
      * Move window to the same screen as main window
      */
-    public moveWindow(windowToMove: BrowserWindow) {
+    public moveWindow(windowToMove: BrowserWindow, fixedYPosition?: number) {
         if (this.mainWindow && windowExists(this.mainWindow)) {
             const display = electron.screen.getDisplayMatching(this.mainWindow.getBounds());
```
KiranNiranjan: @johankwarnmarksymphony I think for screen sharing indicator we should go with focused window rather than `mainWindow` as the RTC can be a popped-out and be placed in a different screen compared with the main window right?

mattias-symphony: I think the designer wants it always on the screen where the main symphony app is

src/app/window-handler.ts
johankwarnmarksymphony: Yes fixed value from the designer Oliver.

src/app/window-handler.ts
johankwarnmarksymphony: @KiranNiranjan Interesting idea, to follow the pop out/focused window.
But let us go with Symphony-app for now as the designer wants

PR #862, opened 2020-01-27
fix: Update Symphony.config

VishwasShashidhar:
Update Symphony.config to revert to `foundation-dev` from `corporate`


VishwasShashidhar:
@johankwarnmarksymphony `foundation-dev` was the default pod that was used on our repo. Somewhere along the lines, it got changed to `corporate`, so, I changed it back to `foundation-dev`

PR #868, opened 2020-02-06
fix: sda-1737: Change the translation of Stop Sharing

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1737 L10N: "Stop sharing" is not aligned in French


VishwasShashidhar:
> Why do we have the definitions in two places?

Idea was to support multiple dialects in the future. Another issue was that on SFE, we sometimes used to get the language value as `en-US`, and, sometimes as `en`, so, for safety, we just have multiple dialect files.

PR #870, opened 2020-02-11
fix: SDA-1743 Use display id to identify screen instead of display position

mattias-symphony:
## Description
Use display id to identify screen instead of display position
[SDA-1743](https://perzoinc.atlassian.net/browse/SDA-1743)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -785,9 +785,7 @@ export class WindowHandler {
                 if (displayId === element.id.toString()) {
                     if (isWindowsOS) {
                         logger.info(`window-handler: element:`, element);
-                        const winX: string = element.bounds.x.toString();
-                        const winY: string = element.bounds.y.toString();
-                        this.execCmd(this.screenShareIndicatorFrameUtil, [ winX, winY ]);
+                        this.execCmd(this.screenShareIndicatorFrameUtil, [ displayId ]);
                     } else {
                         this.createScreenSharingFrameWindow('screen-sharing-frame',
                         element.workArea.width,
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
   },
   "optionalDependencies": {
     "screen-snippet": "git+https://github.com/symphonyoss/ScreenSnippet2.git#v1.0.5",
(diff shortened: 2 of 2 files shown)


src/app/window-handler.ts
```diff
@@ -785,9 +785,7 @@ export class WindowHandler {
                 if (displayId === element.id.toString()) {
                     if (isWindowsOS) {
                         logger.info(`window-handler: element:`, element);
-                        const winX: string = element.bounds.x.toString();
-                        const winY: string = element.bounds.y.toString();
-                        this.execCmd(this.screenShareIndicatorFrameUtil, [ winX, winY ]);
+                        this.execCmd(this.screenShareIndicatorFrameUtil, [ displayId ]);
```
KiranNiranjan: @mattias-symphony There is still an open issue on the `electron` https://github.com/electron/electron/issues/15111 where the `displayId` will be empty in the sources object on `Windows 7`

mattias-symphony: So what should we do, any suggestion?

KiranNiranjan: @mattias-symphony Can we use this https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-monitorfrompoint API to get the display monitor using `x & y`

mattias-symphony: Actually, this whole block of code was already wrapped in `if (displayId !== '')` so it already did not worked on windows 7 - so I think this is fine, we have no regression due to the fix (but possibly another issue, for win7 only, which we might want to handle as a separate issue)

mattias-symphony: > @mattias-symphony Can we use this https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-monitorfrompoint API to get the display monitor using `x & y`

No, x & y will not work because of DPI issues, we can't map accurately from electron x y to windows x y in all scenarios. Using display id is the more stable way to go. Perhaps we could use x y for windows 7 only, if we decide to fix that issue (which I think should be considered separate from this)

PR #872, opened 2020-02-11
fix: SDA-1501 Using ellipsis for notification title overflow

mattias-symphony:
## Description
Adding text-overflow: ellipsis was not enough, so copied all the styles from "message", which was already working

[SDA-1501](https://perzoinc.atlassian.net/browse/SDA-1501)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/src/renderer/styles/notification-comp.less
+++ b/src/renderer/styles/notification-comp.less
@@ -62,13 +62,18 @@ body {
 
 .title {
   font-family: sans-serif;
-  font-size: 14px;
   font-weight: 700;
-  color: var(--text-color);
+  width: 100%;
+  overflow-wrap: break-word;
+  font-size: 14px;
+  margin-top: 5px;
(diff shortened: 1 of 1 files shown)


src/renderer/styles/notification-comp.less
```diff
@@ -62,13 +62,18 @@ body {
 
 .title {
   font-family: sans-serif;
-  font-size: 14px;
   font-weight: 700;
-  color: var(--text-color);
+  width: 100%;
+  overflow-wrap: break-word;
+  font-size: 14px;
+  margin-top: 5px;
```
KiranNiranjan: Do we a margin top here? Can you please attach a screenshot of the new notification?

mattias-symphony: probably best to remove that line :)

PR #874, opened 2020-02-13
fix: SDA-1749: Enable code signing for third party binaries

VishwasShashidhar:
## Description
Due to 3rd party binaries not being signed, notarisation was failing. This PR fixes the issue.

## Solution Approach
Sign 3rd party binaries using electron-builder

## Related PRs
N/A


VishwasShashidhar:
> Do we need to include this file in git? `installer/mac/SymphonySettingsPlugin.bundle/Contents/MacOS/SymphonySettingsPlugin`

Yes, it is required while creating .pkg

PR #876, opened 2020-02-18
fix: Sda 1746

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1746 Popout: Pop-outs open maximized (Mac only)

### DIFF
--- a/src/app/child-window-handler.ts
+++ b/src/app/child-window-handler.ts
@@ -153,6 +153,7 @@ export const handleChildWindow = (webContents: WebContents): void => {
             newWinOptions.alwaysOnTop = mainWindow.isAlwaysOnTop();
             newWinOptions.frame = true;
             newWinOptions.winKey = newWinKey;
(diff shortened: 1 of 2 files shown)


src/app/child-window-handler.ts
```diff
@@ -153,6 +153,7 @@ export const handleChildWindow = (webContents: WebContents): void => {
             newWinOptions.alwaysOnTop = mainWindow.isAlwaysOnTop();
             newWinOptions.frame = true;
             newWinOptions.winKey = newWinKey;
+            newWinOptions.fullscreenable = false;
```
KiranNiranjan: @johankwarnmarksymphony This should be just `fullscreen`. setting `fullscreenable: false` pop-out cannot be made fullscreen
```suggestion
            newWinOptions.fullscreen = false;
```

johankwarnmarksymphony: Great catch I will update

PR #882, opened 2020-03-02
fix: SDA-1717: Screen sharing indicator is not removed when SDA is reloaded

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1717 RTC: Screen sharing indicator is not removed when SDA is reloaded


src/app/window-handler.ts
```diff
@@ -918,6 +918,26 @@ export class WindowHandler {
...
+     * Executes the given command via a child process
+     *
+     * @param util {string}
+     * @param utilArgs {ReadonlyArray<string>}
+     */
+    public execCmd(util: string, utilArgs: ReadonlyArray<string>): Promise<ChildProcess> {
+        logger.info(`window handler: execCmd: util: ${util} utilArgs: ${utilArgs}`);
+        return new Promise<ChildProcess>((resolve, reject) => {
+            return execFile(util, utilArgs, (error: ExecException | null) => {
+                logger.info(`window handler: execCmd: error: ${error}`);
```
mattias-symphony: Avoid printing the error when operation successful:
```suggestion
                if (error) {
                    logger.info(`window handler: execCmd: error: ${error}`);
                }
```

src/app/window-handler.ts
```diff
@@ -918,6 +918,28 @@ export class WindowHandler {
...
+        logger.info(`window handler: execCmd: util: ${util} utilArgs: ${utilArgs}`);
+        return new Promise<ChildProcess>((resolve, reject) => {
+            return execFile(util, utilArgs, (error: ExecException | null) => {
+                if (error) {
+                    logger.info(`window handler: execCmd: error: ${error}`);
+                }
+                if (error && error.killed) {
+                    // processs was killed, just resolve with no data.
+                    return reject(error);
+                }
```
VishwasShashidhar: we should add a comment explaining why we have separated out the `if` statements.

PR #865, opened 2020-01-29
[WIP] fix: SDA-1567: add menu item for managing GPU

VishwasShashidhar:
## Description
Currently, disabling GPU requires users to change the config file. This PR addresses that by adding a menu item for disabling GPU.
[SDA-1567](https://perzoinc.atlassian.net/browse/SDA-1567)

## Solution Approach
Add menu item for managing GPU

## Related PRs
N/A

PR #884, opened 2020-03-03
fix: [SDA-1567] add menu item for disable gpu

VishwasShashidhar:
## Description
To allow users to disable gpu, we create a menu item
[SDA-1567](https://perzoinc.atlassian.net/browse/SDA-1567)

## Solution Approach
Add menu item
Screenshot 2020-03-03 at 17 19 03
Screenshot 2020-03-03 at 17 19 18

## Related PRs
https://github.com/symphonyoss/SymphonyElectron/pull/865/files


package.json
```diff
@@ -108,7 +108,7 @@
     "browserify": "16.2.3",
     "cross-env": "5.2.0",
     "del": "3.0.0",
-    "electron": "6.1.5",
+    "electron": "6.1.7",
```
mattias-symphony: Why do we need to update the electron version for this?

VishwasShashidhar: You are right, it's not relevant for this PR. But, since `6.1.7` has some important fixes, I thought of bundling it with this in the original PR.

@johankwarnmarksymphony Should I raise a separate PR for the upgrade or is it fine here?

VishwasShashidhar: I've reverted back to `6.1.5` as it makes sense to not bundle an electron upgrade with this PR.

Thanks for pointing out @mattias-symphony :)

src/locale/en-US.json
```diff
@@ -125,7 +128,7 @@
   "Quit Symphony": "Quit Symphony",
   "Redo": "Redo",
   "Refresh app when idle": "Refresh app when idle",
-  "Relaunch": "Relaunch",
```
KiranNiranjan: @VishwasShashidhar we need this for https://github.com/symphonyoss/SymphonyElectron/pull/884/files#diff-aed34bfdf22162ca801ca8365863501aR159

VishwasShashidhar: Done @KiranNiranjan, fixed.

PR #907, opened 2020-03-11
fix: SDA-1857 for us keyboard we should also zoom in for command and =

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1857 Shortcuts: Zoom in hot keys not functional (Mac only)

Add so US keyboard can zoom in with "Command" and "="

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -1019,7 +1019,9 @@ export class WindowHandler {
         if (!focusedWindow || !windowExists(focusedWindow)) {
             return;
         }
-        focusedWindow.webContents.zoomLevel += 0.5;
+        // electron/lib/browser/api/menu-item-roles.js row 159
+        const zoomLevelConstantFromElectron = 0.5;
+        focusedWindow.webContents.zoomLevel += zoomLevelConstantFromElectron;
     }
 
     /**
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -964,6 +964,11 @@ export class WindowHandler {
     private registerGlobalShortcuts(): void {
         logger.info(`window-handler: registering global shortcuts!`);
         globalShortcut.register(isMac ? 'Cmd+Alt+I' : 'Ctrl+Shift+I', this.onRegisterDevtools);
+
+        if (isMac) {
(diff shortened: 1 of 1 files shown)


src/app/window-handler.ts
```diff
@@ -1006,6 +1011,17 @@ export class WindowHandler {
         reloadWindow(focusedWindow as ICustomBrowserWindow);
     }
 
+    /**
+     * Zoom in
+     */
+    private onZoomIn(): void {
```
KiranNiranjan: @johankwarnmarksymphony Is this a workaround? do we have an open issue in electron? if so, can you please link to this PR? Thanks

src/app/window-handler.ts
```diff
@@ -1006,6 +1011,17 @@ export class WindowHandler {
...
+        const focusedWindow = BrowserWindow.getFocusedWindow();
+        if (!focusedWindow || !windowExists(focusedWindow)) {
+            return;
+        }
+        focusedWindow.webContents.zoomLevel += 0.5;
```
mattias-symphony: Use a named constant for this? And why 0.5 specifically?

VishwasShashidhar:
@johankwarnmarksymphony We can use [this solution](https://github.com/electron/electron/issues/15496#issuecomment-450685872) instead of writing custom code.

There's an open issue in Electron for this issue @KiranNiranjan , check the link above.

src/app/window-handler.ts
johankwarnmarksymphony: Yes, we are already using the accelerator
zoomIn: 'Ctrl+=',
This accelerator works on windows but not on mac.
So on mac we need to listen for Command/= too

src/app/window-handler.ts
johankwarnmarksymphony: From Electron source code
zoomin: {
    label: 'Zoom In',
    accelerator: 'CommandOrControl+Plus',
    nonNativeMacOSRole: true,
    webContentsMethod: (webContents) => {
      webContents.zoomLevel += 0.5
    }
  },

So we zoom with same level as Command/+

johankwarnmarksymphony:
Better solution will come from @KiranNiranjan

PR #936, opened 2020-03-21
fix: Sda 1899

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1899 Add shortcuts between 1.5, 2.0-stable and 2.0-daily


src/app/window-handler.ts
```diff
@@ -1035,6 +1042,114 @@ export class WindowHandler {
...
     }
 
+    /**
+     * HACK SWITCH to Client 1.5
+     */
+    private async onClient1_5(): Promise <void> {
+        logger.info('window handler: go to Client 1.5');
+        logger.info('this.globalConfig.url: ' + config.getGlobalConfigFields([ 'url']));
+
+        if (config.getGlobalConfigFields([ 'url']).url && config.getGlobalConfigFields([ 'url']).url.startsWith('https://corporate.symphony.com')) {
```
KiranNiranjan: Please move this to a const

src/app/window-handler.ts
```diff
@@ -1035,6 +1042,114 @@ export class WindowHandler {
...
+            logger.info('window handler: switchClient: ' + switchClient);
+            if (switchClient === 'client1_5' ) {
```
KiranNiranjan: I think it's better to rename as currentClient

src/app/window-handler.ts
```diff
@@ -1035,6 +1042,114 @@ export class WindowHandler {
...
+            }
+
+            switchClient = 'client1_5';
+
+            const focusedWindow = BrowserWindow.getFocusedWindow();
+
+            const dogfoodUrl = `https://corporate.symphony.com/`;
+            if (focusedWindow && windowExists(focusedWindow)) {
+                await focusedWindow.loadURL(dogfoodUrl);
+                reloadWindow(focusedWindow as ICustomBrowserWindow);
```
KiranNiranjan: Why do we need to reload after loadURL here?

src/app/window-handler.ts
```diff
@@ -1035,6 +1042,114 @@ export class WindowHandler {
...
+        }
+    }
+
+    /**
+     * HACK SWITCH to Client Mana-stable
+     */
+    private async onClientManaStable(): Promise <void> {
+        logger.info('window handler: go to Client Mana-stable');
+
+        if (config.getGlobalConfigFields([ 'url']).url && config.getGlobalConfigFields([ 'url']).url.startsWith('https://corporate.symphony.com')) {
```

src/app/window-handler.ts
```diff
@@ -1035,6 +1042,114 @@ export class WindowHandler {
...
+            }
+        } else {
+            logger.info('Switch not support for this POD-url');
+        }
+    }
+
+    /**
+     * HACK SWITCH to Client Mana-daily
+     */
+    private async onClientManaDaily(): Promise <void> {
```
KiranNiranjan: We can completely remove this function and use the above function to load the URL dynamically

src/app/window-handler.ts
```diff
@@ -967,6 +979,15 @@ export class WindowHandler {
         globalShortcut.register(isMac ? 'Cmd+Alt+I' : 'Ctrl+Shift+I', this.onRegisterDevtools);
         globalShortcut.register('CmdOrCtrl+R', this.onReload);
 
+        // Hack to switch between Client 1.5, Mana-stable and Mana-daily
+        if (checkIfCorporate()) {
```
KiranNiranjan:
```suggestion
        if (this.globalConfig.url && this.globalConfig.url.startsWith('https://corporate.symphony.com')) {
```

src/app/window-handler.ts
```diff
@@ -46,6 +46,18 @@ export interface ICustomBrowserWindow extends Electron.BrowserWindow {
...
 
+enum ClientVersionTypes {
+    CLIENT_1_5 = 'client_1_5',
+    CLIENT_MANA_STABLE = 'client_mana_stable',
+    CLIENT_MANA_DAILY = 'client_mana_daily',
+}
+
+let currentClient = ClientVersionTypes.CLIENT_1_5;
+
+const checkIfCorporate = () => {
```
KiranNiranjan: we can remove this now

src/app/window-handler.ts
johankwarnmarksymphony: done

src/app/window-handler.ts
johankwarnmarksymphony: done

src/app/window-handler.ts
johankwarnmarksymphony: done

PR #938, opened 2020-03-23
fix: SDA-1268 (Workaround to exit fullscreen when notification is clicked)[RTC]

KiranNiranjan:
## Description
Workaround to fix an issue with exiting fullscreen when in an active RTC session
[SDA-1268](https://perzoinc.atlassian.net/browse/SDA-1268)

## Solution Approach
- exit HTML fullscreen whenever the notifications are clicked


KiranNiranjan:
@johankwarnmarksymphony @mattias-symphony can you review this please?

PR #939, opened 2020-03-23
fix: SDA-1268 (Workaround to exit fullscreen when notification is clicked)[RTC]

KiranNiranjan:


KiranNiranjan:
@VishwasShashidhar @mattias-symphony can you review this please?

PR #944, opened 2020-03-26
fix: SDA-1895 (Escape backslash characters for autoLaunchPath)

KiranNiranjan:
## Description
Escape backslash characters for autoLaunchPath
[SDA-1895](https://perzoinc.atlassian.net/browse/SDA-1895)

## Solution Approach
 - Replace multiple escaped character with a single escape character

## Related PRs

branch | PR
------ | ------
SymphonyElectron | [link](https://github.com/symphonyoss/SymphonyElectron/pull/945)

### DIFF
--- a/src/app/main-api-handler.ts
+++ b/src/app/main-api-handler.ts
@@ -171,6 +171,9 @@ ipcMain.on(apiName.symphonyApi, async (event: Electron.IpcMainEvent, arg: IApiAr
             break;
         case apiCmds.setCloudConfig:
             const { podLevelEntitlements, acpFeatureLevelEntitlements, pmpEntitlements, ...rest } = arg.cloudConfig as ICloudConfig;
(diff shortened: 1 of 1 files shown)


src/app/main-api-handler.ts
```diff
@@ -171,6 +171,9 @@ ipcMain.on(apiName.symphonyApi, async (event: Electron.IpcMainEvent, arg: IApiAr
             break;
         case apiCmds.setCloudConfig:
             const { podLevelEntitlements, acpFeatureLevelEntitlements, pmpEntitlements, ...rest } = arg.cloudConfig as ICloudConfig;
+            if (podLevelEntitlements && podLevelEntitlements.autoLaunchPath && podLevelEntitlements.autoLaunchPath.match(/\\\\/g)) {
```
VishwasShashidhar: what happens if we have six backslashes instead of four that we have today?

KiranNiranjan: @VishwasShashidhar No matter how many `\\\\` there are in the ACP input this check makes sure that it will replace to single escaped backslash characters

PR #959, opened 2020-03-31
feat: SDA-1932: persist client switch

VishwasShashidhar:
## Description
Add logic to persist client switch between 1.5, 2.0 and 2.0 daily
[SDA-1932](https://perzoinc.atlassian.net/browse/SDA-1932)

## Solution Approach
Add parameter to user config to persist client switch across restarts

## Related PRs
N/A

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -247,7 +247,7 @@ export class WindowHandler {
             logger.info(`window-handler: client switch from config is ${this.config.clientSwitch}`);
 
             const parsedUrl = parse(this.url);
-            if (this.url.indexOf(`https://${parsedUrl.hostname}/client/index.html`) !== -1) {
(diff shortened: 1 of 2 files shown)


src/app/window-handler.ts
```diff
@@ -241,6 +244,13 @@ export class WindowHandler {
             }
             this.url = this.mainWindow.webContents.getURL();
 
+            logger.info(`window-handler: client switch from config is ${this.config.clientSwitch}`);
+
+            const parsedUrl = parse(this.url);
+            if (this.url.indexOf(`https://${parsedUrl.hostname}/client/index.html`) !== -1) {
```
KiranNiranjan: I think we should also add this check `this.url.startsWith('https://corporate.symphony.com')` here as well right?

PR #960, opened 2020-04-02
feat: Measure cpu

johankwarnmarksymphony:
Enable RTC team to measure cpu usage for audio / video calls

### DIFF
--- a/src/common/api-interface.ts
+++ b/src/common/api-interface.ts
@@ -36,6 +36,7 @@ export enum apiCmds {
     getConfigUrl = 'get-config-url',
     registerRestartFloater = 'register-restart-floater',
     setCloudConfig = 'set-cloud-config',
+    getCPUUsage = 'get-cpu-usage',
     checkMediaPermission = 'check-media-permission',
 }
 
@@ -143,6 +144,11 @@ export interface IVersionInfo {
     buildNumber: string;
     apiVer: string;
     searchApiVer: string;
+}
+
+export interface ICPUUsage {
+    percentCPUUsage: number;
+    idleWakeupsPerSecond: number;
 }
 
 export interface IMediaPermission {
--- a/src/demo/index.html
+++ b/src/demo/index.html
@@ -144,6 +144,11 @@
         <button id='bring-to-front-without-focus'>bring to front without focus</button>
 
         <hr>
+        <p>Measure cpu usage:</p>
+        <button id='button-get-cpu-usage'>Get CPU usage</button>
+        <input type="text" id='text-cpu-usage' />
+
+        <hr>
         <p>Check Media Permission:</p>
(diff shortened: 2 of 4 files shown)


src/demo/index.html
```diff
@@ -628,6 +638,29 @@ <h1>Symphony Electron API Demo</h1>
             activityTimer = startActivityInterval();
         };
 
+        /**
+         * Get CPU usage
+         */
+        const cpuMessurement = document.getElementById('button-get-cpu-usage');
```
VishwasShashidhar, line 644: typo: please refactor the variable to `cpuMeasurement`

src/demo/index.html
```diff
@@ -628,6 +638,29 @@ <h1>Symphony Electron API Demo</h1>
...
+
+        cpuMessurement.addEventListener('click', () => {
+            console.log('Start messure CPU !!!');
```
VishwasShashidhar, line 647: typo: please rename `messure` to `measure`

src/common/api-interface.ts
```diff
@@ -36,6 +36,7 @@ export enum apiCmds {
     getConfigUrl = 'get-config-url',
     registerRestartFloater = 'register-restart-floater',
     setCloudConfig = 'set-cloud-config',
+    getCPUUsage = 'get-cpu-usage',
```
mattias-symphony, line 39: name should perhaps be getCpuUsage, instead of getCPUUsage (more conventional to not use upper case for abbreviations in camel case)

src/common/api-interface.ts
```diff
@@ -145,6 +146,11 @@ export interface IVersionInfo {
     searchApiVer: string;
 }
 
+export interface ICPUUsage {
+    percentCPUUsage: number;
```
mattias-symphony, line 150: and same here, percentCpuUsage

src/demo/index.html
mattias-symphony, line 647: spelling, "measure CPU"

src/common/api-interface.ts
johankwarnmarksymphony: Will keep because of https://www.electronjs.org/docs/api/structures/cpu-usage

src/common/api-interface.ts
johankwarnmarksymphony: Will keep because of https://www.electronjs.org/docs/api/structures/cpu-usage

src/demo/index.html
johankwarnmarksymphony: fixed

src/demo/index.html
johankwarnmarksymphony: Fixed

src/demo/index.html
johankwarnmarksymphony: fixed

PR #965, opened 2020-04-06
fix: SDA-1945: unregister global shortcuts for client switch on window blur

VishwasShashidhar:
## Description
Unregister global shortcuts for client switch on window blur
[SDA-1945](https://perzoinc.atlassian.net/browse/SDA-1945)

## Solution Approach
Unregister global shortcuts on window blur and register them again once the window is focused.

## Related PRs
N/A


VishwasShashidhar:
Build failure is because of https://github.com/actions/setup-node/issues/77, hence merging this with an override.

PR #971, opened 2020-04-14
fix: SDA-1956: RTC: Missing red border when screen sharing whole screen

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/SDA-1956 RTC: Missing red border when screen sharing whole screen

display_id is empty so we don't know what to share. In this case we will share screen 0.

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -828,9 +828,9 @@ export class WindowHandler {
             });
         }
 
+        logger.info('window-handler: createScreenSharingIndicatorWindow, displayId: ' + displayId);
         if (displayId !== '') {
             const displays = electron.screen.getAllDisplays();
-            logger.info('window-handler: createScreenSharingIndicatorWindow, displayId: ' + displayId);
             displays.forEach((element) => {
                 logger.info('window-handler: element.id.toString(): ' + element.id.toString());
                 if (displayId === element.id.toString()) {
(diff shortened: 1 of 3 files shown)


src/app/window-handler.ts
```diff
@@ -639,6 +639,13 @@ export class WindowHandler {
                     } else if (isMac && type === 'screen') {
                         const dispId = source.id.split(':')[1];
                         this.execCmd(this.screenShareIndicatorFrameUtil, [ dispId ]);
+                    } else if (isWindowsOS && type === 'screen') {
+                        logger.info('window-handler: source.display_id: ' + source.display_id);
```
VishwasShashidhar, line 643: please consider using String interpolation for log statements with variable values.

src/app/window-handler.ts
```diff
@@ -828,23 +835,22 @@ export class WindowHandler {
...
-                    } else if (isLinux) {
-                        this.createScreenSharingFrameWindow('screen-sharing-frame',
-                        element.workArea.width,
-                        element.workArea.height,
-                        element.workArea.x,
-                        element.workArea.y);
+            if (isLinux) {
+                const displays = electron.screen.getAllDisplays();
+                displays.forEach((element) => {
+                    logger.info('window-handler: element.id.toString(): ' + element.id.toString());
```
VishwasShashidhar, line 843: please consider using String interpolation for log statements with variable values.

PR #973, opened 2020-04-21
feat: SDA-1818: simplify macOS installer

VishwasShashidhar:
## Description
- Remove the pod settings plugin
- Set defaults in postinstall.sh
- Set default url to my.symphony.com in postinstall.sh
[SDA-1818](https://perzoinc.atlassian.net/browse/SDA-1818)

## Solution Approach
- Remove source code related to pod settings plugin
- Set default pod url to public pod

PR #977, opened 2020-04-24
feat: SDA-1983: implement simplified installer window on SDA

VishwasShashidhar:
## Description
Implement simplified installer window on SDA to allow users to set pod url upon first launch.
See [SDA-1983](https://perzoinc.atlassian.net/browse/SDA-1983) for more details.

## Solution Approach
See [SDA-1983](https://perzoinc.atlassian.net/browse/SDA-1983) for more details.

## Related PRs
#973


src/app/window-handler.ts
```diff
@@ -160,12 +164,28 @@ export class WindowHandler {
...
         logger.info(`window-handler: initialized spellchecker module with locale ${this.spellchecker.locale}`);
 
         logger.info('window-handler: createApplication mainWinPos: ' + JSON.stringify(this.config.mainWinPos));
 
+        let {isFullScreen, isMaximized} = this.config.mainWinPos ? this.config.mainWinPos : {isFullScreen: false, isMaximized: false};
+
+        this.url = WindowHandler.getValidUrl(this.userConfig.url ? this.userConfig.url : this.globalConfig.url);
+        logger.info(`window-handler: setting url ${this.url} from config file!`);
+
+        if (this.globalConfig.url.startsWith('https://my.symphony.com') && !this.userConfig.url) {
```
johankwarnmarksymphony, line 178: Will this not show the dialog every time for a user that want to use my.symphony.com

Don't you need a flag e.g. showWelcome

VishwasShashidhar: No @johankwarnmarksymphony. It shows the dialog only if the url is not set in user config file.

And, once a user sets the url, we update that in the user config file, so, this condition won't come into picture after that.

```
ipcMain.on('set-pod-url', async (_event, newPodUrl: string) => {
             await config.updateUserConfig({url: newPodUrl, mainWinPos: { ...this.mainWindow!.getPosition(), ...{ height: 900, width: 900 } } });
             app.relaunch();
             app.exit();
         });
```

src/app/config-handler.ts
```diff
@@ -324,7 +324,9 @@ class Config {
             await app.whenReady();
             await this.readGlobalConfig();
             logger.info(`config-handler: user config doesn't exist! will create new one and update config`);
-            await this.updateUserConfig({ configVersion: app.getVersion().toString(), buildNumber, ...this.globalConfig } as IConfig);
+            const newUserConfig = this.globalConfig as IConfig;
+            delete newUserConfig.url;
```
KiranNiranjan, line 328: This will also remove the URL field from `this.globalConfig`

VishwasShashidhar: You are right, fixed and pushed a new commit.

src/app/config-handler.ts
```diff
@@ -324,7 +324,7 @@ class Config {
             await app.whenReady();
             await this.readGlobalConfig();
             logger.info(`config-handler: user config doesn't exist! will create new one and update config`);
-            const newUserConfig = this.globalConfig as IConfig;
+            const newUserConfig = JSON.parse(JSON.stringify(this.globalConfig)) as IConfig;
```
KiranNiranjan, line 327:
```suggestion
            const { url, ...rest } = this.globalConfig;
```

PR #993, opened 2020-05-12
fix: SDA-1738 RTC diagnotic start/close screen share quickly

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-1738 RTC: Shared screen red border still shows after RTC diagnostics is completed

RTC diagnostic start/close screen share quickly add timeout for 300ms

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -510,7 +510,7 @@ export class WindowHandler {
                     }
                 }
                 if (isWindowsOS || isMac) {
-                    this.execCmd(this.screenShareIndicatorFrameUtil, []);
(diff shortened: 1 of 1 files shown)


src/app/window-handler.ts
```diff
@@ -512,7 +512,7 @@ export class WindowHandler {
                     }
                 }
                 if (isWindowsOS || isMac) {
-                    this.execCmd(this.screenShareIndicatorFrameUtil, []);
+                    setTimeout(() => this.execCmd(this.screenShareIndicatorFrameUtil, []), 300);
```
VishwasShashidhar, line 515: Please move the value to a constant.

PR #997, opened 2020-05-15
feat: SDA-1991 (Redesign Notifications)

KiranNiranjan:
## Description
Redesign Notification
[SDA-1991](https://perzoinc.atlassian.net/browse/SDA-1991)

## Solution Approach

- Redesign notifications for `mana` and `1.5`
- Support custom notification for Mac 🎉 

#### Mana
Screenshot 2020-05-14 at 5 48 08 PM
Screenshot 2020-05-14 at 5 49 09 PM
Screenshot 2020-05-15 at 2 31 17 PM

#### 1.5
Screenshot 2020-05-15 at 12 57 56 PM

### DIFF
--- a/src/common/api-interface.ts
+++ b/src/common/api-interface.ts
@@ -111,6 +111,8 @@ export enum KeyCodes {
     Alt = 18,
 }
 
+type Theme = '' | 'light' | 'dark';
+
 /**
  * Notification
  */
@@ -125,6 +127,8 @@ export interface INotificationData {
     sticky: boolean;
     company: string;
     displayTime: number;
+    isExternal: boolean;
+    theme: Theme;
 }
 
 export enum NotificationActions {
--- a/src/renderer/components/notification-comp.tsx
+++ b/src/renderer/components/notification-comp.tsx
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { i18n } from '../../common/i18n-preload';
 
 const whiteColorRegExp = new RegExp(/^(?:white|#fff(?:fff)?|rgba?\(\s*255\s*,\s*255\s*,\s*255\s*(?:,\s*1\s*)?\))$/i);
-const darkTheme = [ '#b5616a', '#ab8ead', '#ebc875', '#a3be77', '#58c6ff', '#ebab58' ];
+const darkTheme = [ '#e23030', '#b5616a', '#ab8ead', '#ebc875', '#a3be77', '#58c6ff', '#ebab58' ];
 type Theme = '' | 'light' | 'dark';
 
 interface IState {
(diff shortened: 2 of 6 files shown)


VishwasShashidhar:
This looks so much better!

src/common/api-interface.ts
```diff
@@ -111,6 +111,8 @@ export enum KeyCodes {
     Alt = 18,
 }
 
+type Theme = '' | 'light' | 'dark';
```
VishwasShashidhar, line 114: do we really need an empty string here?

src/renderer/components/notification-comp.tsx
```diff
@@ -1,10 +1,11 @@
-import classNames from 'classnames';
 import { ipcRenderer } from 'electron';
 import * as React from 'react';
 
 import { i18n } from '../../common/i18n-preload';
 
 const whiteColorRegExp = new RegExp(/^(?:white|#fff(?:fff)?|rgba?\(\s*255\s*,\s*255\s*,\s*255\s*(?:,\s*1\s*)?\))$/i);
+const darkTheme = [ '#b5616a', '#ab8ead', '#ebc875', '#a3be77', '#58c6ff', '#ebab58' ];
+type Theme = '' | 'light' | 'dark';
```
VishwasShashidhar, line 8: do we really need an empty string here?

src/renderer/notification-handler.ts
```diff
@@ -64,8 +64,8 @@ export default class NotificationHandler {
      * Initializes / resets the notification positional values
      */
     public setupNotificationPosition() {
-        // This feature only applies to windows
-        if (isMac || isLinux || !app.isReady()) {
+        // This feature only applies to windows & mac
+        if (isLinux || !app.isReady()) {
```
VishwasShashidhar, line 68: the code you've written should actually support linux as well right?

VishwasShashidhar:
@KiranNiranjan Can we slide out notifications on macOS custom notifications? If not, can you check if it is possible to implement it?

KiranNiranjan:
> @KiranNiranjan Can we slide out notifications on macOS custom notifications? If not, can you check if it is possible to implement it?

@VishwasShashidhar There are some APIs to support this, but not a straight forward one. There will be some manual calculation that needs to be done (including trackpad monitor for animated close)

src/renderer/components/notification-comp.tsx
KiranNiranjan: Yes, this is because of the `1.5` dependency

src/renderer/notification-handler.ts
KiranNiranjan: Sure, will update

PR #1025, opened 2020-06-24
fix: SDA-2179 Ensuring screen share indicator frame is turned off when switching client versions

mattias-symphony:
## Description
Ensuring screen share indicator frame is turned off when switching client versions, both using hotkeys and menu option

[SDA-2197](https://perzoinc.atlassian.net/browse/SDA-2197)

## Solution Approach
Closing the screen share indicator when the mainwindow url change and when hotkeys are invoked

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -1277,6 +1277,7 @@ export class WindowHandler {
                 default:
                     this.url = this.globalConfig.url + `?x-km-csrf-token=${csrfToken}`;
             }
+            this.execCmd(this.screenShareIndicatorFrameUtil, []);
             await this.mainWindow.loadURL(this.url);
         } catch (e) {
             logger.error(`window-handler: failed to switch client because of error ${e}`);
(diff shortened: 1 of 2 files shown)


src/app/window-handler.ts
```diff
@@ -1277,6 +1277,7 @@ export class WindowHandler {
                 default:
                     this.url = this.globalConfig.url + `?x-km-csrf-token=${csrfToken}`;
             }
+            this.execCmd(this.screenShareIndicatorFrameUtil, []);
```
KiranNiranjan, line 1280: I think we need to move all the screen sharing indicator logic to a separate handler file. 

I have included this change in https://perzoinc.atlassian.net/browse/SDA-2066 point 7.

mattias-symphony: yes, that sounds like a reasonable improvement

PR #1041, opened 2020-07-28
chore: Move electron forward to 9.1.1

johankwarnmarksymphony:
Move from Electron 9.0.2 to 9.1.1
https://www.electronjs.org/releases/stable


package.json
```diff
@@ -121,7 +121,7 @@
     "browserify": "16.5.1",
     "cross-env": "5.2.0",
     "del": "3.0.0",
-    "electron": "9.0.2",
+    "electron": "9.1.1",
```
KiranNiranjan, line 124: @johankwarnmarksymphony There is a new release `9.1.2`

KiranNiranjan, line 124:
```suggestion
    "electron": "9.1.2",
```

PR #1044, opened 2020-08-04
fix: Change the title on corporate to indicate stable or daily

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-2343 Show which SFE-lite client the SDA is running


johankwarnmarksymphony:
not needed

PR #1056, opened 2020-08-21
feat: SDA-2388 WixSharp Initial project setup

mattias-symphony:
## Description
Added the WixSharp Toolset to the repo
Initial project setup with all files included, and batch file for running building the MSI
Added shortcuts to desktop and program menu
Added launch condition to require Windows Server 2008 or later

It is only necessary to review 2 files, BuildWixSharpInstaller.bat and Symphony.cs - the rest are just a copy of the latest release of WixSharp Toolset.

[SDA-2388](https://perzoinc.atlassian.net/browse/SDA-2388)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()


installer/win/Symphony.cs
```diff
@@ -0,0 +1,111 @@
...
+        // the file list is rather long and explicit. At some point we might make the `dist` folder match exactly the 
+        // desired contents of installation, and then we can simplify this bit.
+        var project = new ManagedProject(productName,
+            new Dir(@"%ProgramFiles%\" + productName,
+                new File(@"..\..\dist\win-unpacked\Symphony.exe",
+                    // Create two shortcuts to the main Symphony.exe file, one on the desktop and one in the program menu
+                    new FileShortcut(productName, @"%Desktop%") { IconFile = @"..\..\images\icon.ico" },
+                    new FileShortcut(productName, @"%ProgramMenu%") { IconFile = @"..\..\images\icon.ico" }
+                ),
+                new File(@"..\..\dist\win-unpacked\chrome_100_percent.pak"),
```
KiranNiranjan, line 26: @mattias-symphony is there a way to pick all the files from `\win-unpacked\`?

Because when we upgrade to newer versions of the electron there is a chance a binary file being included/removed from this directory, then we would manually come here to include those files?

mattias-symphony: This is the same way as we have with Advanced Installer currently. Files are added manually, and not on a per-directory basis. With the new installer, we could indeed pick all files from win-unpacked if we wanted to, the problem is that we don't want to include all the files in win unpacked (it contains files which we don't want to add to the installer), and also that not all files we want to include are from win-unpacked.

As outlined in the comment in Setup.cs, we could change the output of our build scripts so that the win unpacked (dist) folder mirrors exactly what we want in the installer, and then we can simplify the installer script. This is to be considered a separate Improvement task, should we choose to do it, as it is not something present in current installer.

installer/win/BuildWixSharpInstaller.bat
```diff
@@ -0,0 +1,2 @@
+set WIXSHARP_WIXDIR=.\WixSharpToolSet\Wix_bin\bin
+.\WixSharpToolSet\cscs Symphony.cs
```
VishwasShashidhar, line 2: I think it'd be better to have a new line at the end of a file.

VishwasShashidhar (approved):
@mattias-symphony Please see what I suggested makes sense. If it does, you can make those changes, if not, I think this looks good, we can go ahead.

installer/win/Symphony.cs
VishwasShashidhar: > Yeah, WixSharp is distributed as a .7z compressed archive, so upgrading is just a matter of unzipping it and copying the files.

I agree, we should bundle this with our codebase.

mattias-symphony: Hmm, not sure we can do it from config file without complicating things a lot - I think the better approach is to have a separate step in the build scripts which collates all necessary files, and then we just add *.* with subfolders on that collection of files. The script which collates the files could then be config file driven if we want to. But I think this is a future improvement

installer/win/BuildWixSharpInstaller.bat
mattias-symphony: It's a batch file, it doesn't care :)

installer/win/Symphony.cs
```diff
@@ -0,0 +1,111 @@
...
+//css_ref WixSharp.UI.dll;
+
+using WixSharp;
+
+class Script
+{
+    static public void Main(string[] args)
+    {
+        // The name "Symphony" is used in a lot of places, for paths, shortut names and installer filename, so define it once
+        var productName = "Symphony";
```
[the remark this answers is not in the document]
mattias-symphony: Sure, but why? It is not likely to change often

PR #1061, opened 2020-08-28
fix: sda-2261 translate enable console logs

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-2261 SDA 9.0.0.452 - Localization - Sub -Menu - 'Enable Renderer Logs' is not localized


mattias-symphony (approved):
Why do we have the duplication, specifying everything in two places?

VishwasShashidhar (approved):

PR #1064, opened 2020-09-03
fix: SDA-2279 (Filter out notification windows from desktop capturer)

KiranNiranjan:
## Description
Filter out custom notification window from showing up in the screen share application tab
[SDA-2279](https://perzoinc.atlassian.net/browse/SDA-2279)

## Solution Approach
- Filter out all the window with name `Notification - Symphony`

notes: custom notifications will no longer be displayed in screen picker application tab

### DIFF
--- a/src/renderer/desktop-capturer.ts
+++ b/src/renderer/desktop-capturer.ts
@@ -120,7 +120,9 @@ export const getSource = async (options: ICustomSourcesOptions, callback: Callba
 
     }
 
-    const updatedSources = sources.map((source) => {
+    const updatedSources = sources
+        .filter((source) => isWindowsOS ? source.name !== 'Notification - Symphony' : true)
+        .map((source) => {
         return Object.assign({}, source, {
             thumbnail: source.thumbnail.toDataURL(),
         });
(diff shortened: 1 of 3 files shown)


src/renderer/desktop-capturer.ts
```diff
@@ -120,7 +120,9 @@ export const getSource = async (options: ICustomSourcesOptions, callback: Callba
 
     }
 
-    const updatedSources = sources.map((source) => {
+    const updatedSources = sources
+        .filter((source) => isWindowsOS ? source.name !== NOTIFICATION_WINDOW_TITLE : true)
```
VishwasShashidhar, line 124: we use custom notifications on macOS as well now right? Should you remove the `isWindowsOS` condition here?

VishwasShashidhar (commented):
@KiranNiranjan I just have one comment pertaining to `isWindowsOS` check. Please take a look at it.

src/renderer/desktop-capturer.ts
KiranNiranjan: @VishwasShashidhar MacOS is not considering `notification windows` as applications. so, the issue is only happening on Windows

johankwarnmarksymphony (approved):

keerthi16 (approved):

mattias-symphony (approved):

PR #1067, opened 2020-09-08
feat: SDA-2393 Adding/removing registry keys

mattias-symphony:
## Description
Adding registry key for protocol handlers, and a custom action to remove registry keys on installation (including keys added by SDA, like auto-launch)
[SDA-2393](https://perzoinc.atlassian.net/browse/SDA-2393)

## Solution Approach

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/installer/win/WixSharpInstaller/Symphony.cs
+++ b/installer/win/WixSharpInstaller/Symphony.cs
@@ -11,6 +11,7 @@
 using WixSharp;
 using WixSharp.Forms;
 using Microsoft.Deployment.WindowsInstaller;
+using Microsoft.Win32;
 
 class Script
 {
@@ -18,10 +19,10 @@ class Script
     {
         // The name "Symphony" is used in a lot of places, for paths, shortut names and installer filename, so define it once
         var productName = "Symphony";
-        
+
         // Create a wixsharp project instance and assign the project name to it, and a hierarchy of all files to include
-        // Files are taken from multiple locations, and not all files in each location should be included, which is why 
-        // the file list is rather long and explicit. At some point we might make the `dist` folder match exactly the 
+        // Files are taken from multiple locations, and not all files in each location should be included, which is why
+        // the file list is rather long and explicit. At some point we might make the `dist` folder match exactly the
         // desired contents of installation, and then we can simplify this bit.
         var project = new ManagedProject(productName,
             new Dir(@"%ProgramFiles%\" + productName,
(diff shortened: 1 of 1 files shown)


installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -258,11 +273,68 @@ public static ActionResult UpdateConfig(Session session)
...
     }
 
+    // CleanRegistry custom action
+    [CustomAction]
+    public static ActionResult CleanRegistry(Session session)
+    {
+        try
+        {
+            // Remove registry keys added for auto-launch
+
```
VishwasShashidhar, line 290: Is this blank line by mistake?

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -258,11 +273,68 @@ public static ActionResult UpdateConfig(Session session)
...
+            using( var key = Registry.CurrentUser.OpenSubKey(@"Software\Microsoft\Windows\CurrentVersion\Run", true) )
+            {
+                if (key != null)
+                {
+                    key.DeleteValue("Symphony", false);
```
VishwasShashidhar, line 295: what is false in this case? Can we move it to a variable and use the variable here so the code is self explanatory?

Same for all the following occurrences.

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -258,11 +273,68 @@ public static ActionResult UpdateConfig(Session session)
...
+                if (key != null)
+                {
+                    key.DeleteSubKeyTree("symphony", false);
+                }
+            }
+        }
+        catch (System.Exception e)
+        {
+            session.Log("Error executing CleanRegistry: " + e.ToString() );
+            return ActionResult.Success;
```
VishwasShashidhar, line 336: can we use finally and return the result in one step?

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony: No

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony: it is just the API for C# DeleteValue - second parameter is throwOnMissingKey or something like that. But I don't think we need a variable for this, I think it is a case where you would look up the API function anyway, as it is a Win32 API function and there are usually subtleties there if you want to understand what is happening

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony: yeah, but I think it is cleaner to do it like this, and more in line with the other custom actions which follows this pattern (as they can return ActionResult.Error)

VishwasShashidhar: ok. It's a different way of coding anyway.

VishwasShashidhar (approved):

johankwarnmarksymphony (approved):

PR #1076, opened 2020-09-16
fix: SDA-2465 Clamping the display id to ensure we are not accessing outside the array

mattias-symphony:
## Description
[SDA-2465](https://perzoinc.atlassian.net/browse/SDA-2465)

## Solution Approach
To ensure we are not reading outside of the array, we clamp the value to not be outside of valid indices for the array.

## Related PRs

branch | PR
------ | ------
other_pr_dev | [link]()

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -803,10 +803,12 @@ export class WindowHandler {
                             this.execCmd(this.screenShareIndicatorFrameUtil, [ source.display_id ]);
                         } else {
                             const dispId = source.id.split(':')[1];
+                            const clampedDispId = Math.min(dispId, displays.length - 1);
                             const keyId = 'id';
                             logger.info('window-utils: dispId: ' + dispId);
-                            logger.info('window-utils: displays [' + dispId + '] [id]: ' + displays [dispId] [ keyId ]);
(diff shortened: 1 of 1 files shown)


psjostrom (approved):

VishwasShashidhar:
@mattias-symphony Please describe the solution in the description. It'll help in the future if someone looks at the code to understand what clamping does exactly.

johankwarnmarksymphony (approved):

mattias-symphony:
> @mattias-symphony Please describe the solution in the description. It'll help in the future if someone looks at the code to understand what clamping does exactly.

what more description could you possibly want for a small thing like this? The problem is in the ticket, the commit message tells what was done. How would you have described it without being redundant?

VishwasShashidhar:
> what more description could you possibly want for a small thing like this? The problem is in the ticket, the commit message tells what was done. How would you have described it without being redundant?

Adding a code comment where you are writing the logic helps someone else understand what you are doing there. For example, you are taking the minimum of dispID and displays.length below

`Math.min(dispId, displays.length - 1);`

Why are we taking this approach?

And, you'll need to fill in the PR description appropriately as per the process. I understand the JIRA ticket describes the details, but, this is an open source project and some people don't have access to JIRA, and, that is the reason we have this template.

johankwarnmarksymphony:
CLA stage is not working

PR #1079, opened 2020-09-18
feat: SDA-2395 Upgrading existing Symphony

mattias-symphony:
## Description
Handle upgrading of Symphony properly, by setting product GUID and upgrade code appropriately, and handling closing down of the app if it is running when installing
[SDA-2395](https://perzoinc.atlassian.net/browse/SDA-2395)

## Solution Approach
For the close dialog, it is not possible to close the app when the user press "Close Symphony" button, but instead the stopping of the app happens just before it starts installing files.

## Related PRs
N/A

notes: address use case of upgrading an existing version of SDA in the installer

### DIFF
--- a/installer/win/WixSharpInstaller/CloseDlg.cs
+++ b/installer/win/WixSharpInstaller/CloseDlg.cs
@@ -13,16 +13,16 @@ namespace Symphony
 
         void dialog_Load(object sender, System.EventArgs e)
         {
-			// Detect if Symphony is running
+            // Detect if Symphony is running
             bool isRunning = System.Diagnostics.Process.GetProcessesByName("Symphony").Length > 0;
             if (isRunning)
-			{
-				// If it is running, disable the "next" button
-				this.next.Enabled = false;
-			}
-			else
             {
-				// If it is not running, skip this dialog
+                // If it is running, disable the "next" button
+                this.next.Enabled = false;
+            }
+            else
+            {
+                // If it is not running, skip this dialog
                 Shell.GoNext();
             }
 
@@ -33,8 +33,8 @@ namespace Symphony
 
         void closeButton_Click(object sender, System.EventArgs e)
         {
-			// The "Close Symphony" button is just to get users consent to close the app.
-			// Actually closing the app will be done later in the flow.
+            // The "Close Symphony" button is just to get users consent to close the app.
+            // Actually closing the app will be done later in the flow.
             Shell.GoNext();
         }
 
--- a/installer/win/WixSharpInstaller/CloseDlg.cs
+++ b/installer/win/WixSharpInstaller/CloseDlg.cs
@@ -1,0 +1,57 @@
+using WixSharp;
+using System.Drawing;
+
+namespace Symphony
+{
+    public partial class CloseDlg : WixSharp.UI.Forms.ManagedForm, IManagedDialog
+    {
+
+        public CloseDlg()
+        {
+            InitializeComponent();
+        }
+
+        void dialog_Load(object sender, System.EventArgs e)
+        {
+			// Detect if Symphony is running
+            bool isRunning = System.Diagnostics.Process.GetProcessesByName("Symphony").Length > 0;
+            if (isRunning)
+			{
+				// If it is running, disable the "next" button
+				this.next.Enabled = false;
+			}
+			else
+            {
+				// If it is not running, skip this dialog
+                Shell.GoNext();
+            }
+
+            // Populate the dynamic UI elements that can't be set at compile time (background image and
+            // the enabled/disabled state of the `Close Symphony` button)
+            this.backgroundPanel.BackgroundImage = Runtime.Session.GetResourceBitmap("WixUI_Bmp_Dialog");
+        }
+
+        void closeButton_Click(object sender, System.EventArgs e)
+        {
+			// The "Close Symphony" button is just to get users consent to close the app.
+			// Actually closing the app will be done later in the flow.
+            Shell.GoNext();
+        }
+
+        void back_Click(object sender, System.EventArgs e)
+        {
+            Shell.GoPrev();
+        }
+
+        void next_Click(object sender, System.EventArgs e)
+        {
+            Shell.GoNext();
+        }
+
+        void cancel_Click(object sender, System.EventArgs e)
+        {
+            Shell.Cancel();
+        }
+
+    }
+}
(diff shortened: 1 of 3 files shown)


KiranNiranjan (approved):

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
+using WixSharp;
+using System.Drawing;
+
+namespace Symphony
+{
+    public partial class CloseDlg : WixSharp.UI.Forms.ManagedForm, IManagedDialog
```
VishwasShashidhar, line 6: can we refactor this to be called `CloseDialog` instead of the short form?

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+
+        public CloseDlg()
+        {
+            InitializeComponent();
+        }
+
+        void dialog_Load(object sender, System.EventArgs e)
```
VishwasShashidhar, line 14: is this an overridden method? We are following both the snake & camel case here. Can we keep it consistent and follow the proper casing as that is the standard for C# ?

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
...
+                // If it is not running, skip this dialog
+                Shell.GoNext();
+            }
+
+            // Populate the dynamic UI elements that can't be set at compile time (background image and
+            // the enabled/disabled state of the `Close Symphony` button)
+            this.backgroundPanel.BackgroundImage = Runtime.Session.GetResourceBitmap("WixUI_Bmp_Dialog");
+        }
+
+        void closeButton_Click(object sender, System.EventArgs e)
```
VishwasShashidhar, line 34: same as above.

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            // The "Close Symphony" button is just to get users consent to close the app.
+            // Actually closing the app will be done later in the flow.
+            Shell.GoNext();
+        }
+
+        void back_Click(object sender, System.EventArgs e)
```
VishwasShashidhar, line 41: same as above.

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Shell.GoPrev();
+        }
+
+        void next_Click(object sender, System.EventArgs e)
```
VishwasShashidhar, line 46: same as above.

installer/win/WixSharpInstaller/CloseDlg.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Shell.GoNext();
+        }
+
+        void cancel_Click(object sender, System.EventArgs e)
```
VishwasShashidhar, line 51: same as above.

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -109,16 +111,30 @@ static public void Main(string[] args)
...
+        // More details can be found in this stackoverflow post:
+        //      https://stackoverflow.com/a/26344742
+        project.GUID = System.Guid.NewGuid();
+        project.UpgradeCode = new System.Guid("{a9b448c9-f065-41a4-87c3-da527c6a389b}");
+
+        // Don't allow installation of earlier versions, but do allow installing the same version as is already installed.
+        project.MajorUpgrade = new MajorUpgrade();
+        project.MajorUpgrade.AllowDowngrades = false;
+        project.MajorUpgrade.AllowSameVersionUpgrades = true;
+        project.MajorUpgrade.DowngradeErrorMessage = "A more recent version of Symphony is already installed on this computer.";
```
VishwasShashidhar, line 136: So, in this case, if a user wants to downgrade, they'll need to uninstall the app first and then install the older version. Is that correct?

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -183,18 +199,53 @@ static public void Main(string[] args)
...
 
+        project.Load += project_Load;
 
+        project.Platform = Platform.x64;
 
         // Generate an MSI from all settings done above
         Compiler.BuildMsi(project);
     }
+
+    static void project_Load(SetupEventArgs e)
```
VishwasShashidhar, line 218: please use proper casing for the method name.

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -183,18 +199,53 @@ static public void Main(string[] args)
...
+                        p.Kill();
+                        p.WaitForExit();
+                    }
+                }
+            });
+        }
+        catch (System.ComponentModel.Win32Exception ex)
+        {
+            // We always seem to get this specific exception triggered, but the application still close down correctly.
+            // The exception descriptionis "Only part of a ReadProcessMemory or WriteProcessMemory request was completed".
```
VishwasShashidhar, line 237: spacing missing between `description` and `is`

VishwasShashidhar (changes requested):

psjostrom (approved):
Looks good, I agree with Vish though about consistensy with snake- and camelcase.

installer/win/WixSharpInstaller/CloseDlg.cs
mattias-symphony: Keeping it consistent with the already added dialogs for now. We can change it, but that would be a separate improvement that we should apply to all dialogs in that case.

installer/win/WixSharpInstaller/CloseDlg.cs
mattias-symphony: It is overridden, we can not change the name

installer/win/WixSharpInstaller/CloseDlg.cs
mattias-symphony: These names are created by winforms

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony: yes, I believe that is how the current one works also

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony: this is how it is done in wixsharp examples, I think it is good to follow that convention.

installer/win/WixSharpInstaller/Symphony.cs
mattias-symphony, line 237:
```suggestion
            // The exception description is "Only part of a ReadProcessMemory or WriteProcessMemory request was completed".
```

VishwasShashidhar (approved):

PR #1082, opened 2020-09-23
fix: SDA-2421: copy config on startup

VishwasShashidhar:
## Description
- Currently, if a user config file is already present, and if any changes are made to any parameter in the global config file, they don't get reflected in the user config file. This can be a problem for IT administrators whose changes to the global config file don't reflect.
- This PR addresses that problem by copying over config values that do not affect user settings but helps IT admins.
[SDA-2421](https://perzoinc.atlassian.net/browse/SDA-2421)

## Solution Approach
- Copy the relevant global config values to user config on startup
- Also format the user config and cloud config files on write

## Related PRs
N/A

notes: copy global config values to user config on startup

### DIFF
--- a/src/app/config-handler.ts
+++ b/src/app/config-handler.ts
@@ -148,6 +148,7 @@ class Config {
         this.userConfig = {};
         this.cloudConfig = {};
         this.filteredCloudConfig = {};
+
         this.readUserConfig();
         this.readGlobalConfig();
         this.readInstallVariant();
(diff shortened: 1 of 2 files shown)


src/app/config-handler.ts
```diff
@@ -283,6 +284,27 @@ class Config {
         await this.updateUserConfig({ buildNumber, installVariant: this.installVariant });
     }
 
+    /**
```
johankwarnmarksymphony, line 287: disableGPU, is user setting in the menu.

If that value is written to user settings, it will be overwritten at every startup.
Is that the case ?

Same goes enableRendererLogs.

What is the priority of the 3 different configs ?
cloud > global > user

We need to think at every parameter, which config has priority.

As user if I do change I expect it to over rule the cloud and global

johankwarnmarksymphony (changes requested):

src/app/config-handler.ts
mattias-symphony: agree, i think it is worth clarifying priorities everywhere they are used, by comment (like: this overrides the param defined in that place, and is overriden by a param defined at this other place)

mattias-symphony (approved):

src/app/config-handler.ts
VishwasShashidhar: @johankwarnmarksymphony Removed disableGpu and enableRendererLogs. Rest of the fields are controlled by admins.

None of the user related fields will be overwritten now.

And, priority is Cloud > User > Global

@mattias-symphony I've added a code comment.

johankwarnmarksymphony (approved):

psjostrom (approved):

KiranNiranjan (approved):

PR #1101, opened 2020-10-21
feat: SDA-2390 Modified build script to call signing script on build machine

mattias-symphony:
## Description
Adding signing of the msi for new installer.
[SDA-2390](https://perzoinc.atlassian.net/browse/SDA-2390)

## Solution Approach
Signing msi by calling a signing.bat script stored locally on build machine

Notes: Modified build script to call signing script on build machine

### DIFF
--- a/scripts/jenkins-win64.bat
+++ b/scripts/jenkins-win64.bat
@@ -98,4 +98,9 @@ copy "%AIP%-SetupFiles\%AIP%.msi" "%targetsDir%\%archiveName%.msi"
 
 echo Building new installer with wix#
 call "BuildWixSharpInstaller.bat"
+
+if EXIST c:\electron-installer\signing.bat (
+	call c:\electron-installer\signing.bat
+)
+
 copy "WixSharpInstaller\Symphony.msi" "%targetsDir%\Experimental-%archiveName%.msi"
--- a/scripts/jenkins-win64.bat
+++ b/scripts/jenkins-win64.bat
@@ -99,8 +99,10 @@ copy "%AIP%-SetupFiles\%AIP%.msi" "%targetsDir%\%archiveName%.msi"
 echo Building new installer with wix#
 call "BuildWixSharpInstaller.bat"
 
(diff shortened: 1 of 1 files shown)


KiranNiranjan (approved):

scripts/jenkins-win64.bat
```diff
@@ -98,4 +98,9 @@ copy "%AIP%-SetupFiles\%AIP%.msi" "%targetsDir%\%archiveName%.msi"
 
 echo Building new installer with wix#
 call "BuildWixSharpInstaller.bat"
+
+if EXIST c:\electron-installer\signing.bat (
+	call c:\electron-installer\signing.bat
+)
+
```
johankwarnmarksymphony, line 105: If it does not exist I think you should fail the build.
As it is now, we will just continue and release an unsigned build

johankwarnmarksymphony (changes requested):

scripts/jenkins-win64.bat
mattias-symphony: Good point. But the check was added so that you will be able to run the build script, including making the installer, on your local machine, where there is no certificate. But I guess ensuring that he build is always signed is more important?

VishwasShashidhar: I agree with @mattias-symphony here. When we build on Jenkins, we know we'll have a signed build, so, there's a guarantee there. In any case, we don't distribute builds building locally but, it'll be useful for testing.

VishwasShashidhar (approved):

scripts/jenkins-win64.bat
johankwarnmarksymphony: Team decide we will exit if the signing tool is missing.

For locally run you will have the installer, but you will exit != 0

mattias-symphony: Changed the behaviour as agreed

johankwarnmarksymphony (approved):

PR #1102, opened 2020-10-30
feat: SDA-2593: new design for a simplified installer flow

mattias-symphony:
## Description
There is a new design for a simplified installer flow
[SDA-2593](https://perzoinc.atlassian.net/browse/SDA-2593)

## Solution Approach
Implemented new flow as designed

Notes: New design for a simplified installer flow


KiranNiranjan (approved):

installer/win/WixSharpInstaller/CloseDialog.cs
```diff
@@ -32,22 +29,21 @@ void closeButton_Click(object sender, System.EventArgs e)
...
         void next_Click(object sender, System.EventArgs e)
         {
-            Shell.GoTo(InstallDirIndex);
+            Shell.GoNext();
         }
 
         void cancel_Click(object sender, System.EventArgs e)
         {
-            Shell.Cancel();
+            if( System.Windows.Forms.MessageBox.Show("Are you sure you want to cancel Symphony installation?",
```
VishwasShashidhar, line 42: we should perhaps get this string from a strings file to support internationalisation in the future.

installer/win/WixSharpInstaller/ProgressDialog.cs
```diff
@@ -0,0 +1,218 @@
...
+                                   messageRecord[1]    "InstallFiles"
+                                   messageRecord[2]    "Copying new files"
+                                   messageRecord[3]    "File: [1],  Directory: [9],  Size: [6]"
+
+                                   messageRecord[0]    "Action 23:15:21: [1]. [2]"
+                                   messageRecord[1]    "RegisterUser"
+                                   messageRecord[2]    "Registering user"
+                                   messageRecord[3]    "[1]"
+
+                                */
```
VishwasShashidhar, line 136: I think we should remove this commented out code.

installer/win/WixSharpInstaller/ProgressDialog.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        /// <summary>
+        /// Handles the Click event of the cancel control.
+        /// </summary>
+        /// <param name="sender">The source of the event.</param>
+        /// <param name="e">The <see cref="System.EventArgs" /> instance containing the event data.</param>
+        void cancel_Click(object sender, EventArgs e)
+        {
+            showingCancelConfirmation = true;
+            if (System.Windows.Forms.MessageBox.Show("Are you sure you want to cancel Symphony installation?",
```
VishwasShashidhar, line 206: we should get this string from a strings file for internationalisation support in the future

installer/win/WixSharpInstaller/WelcomeDialog.cs
```diff
@@ -55,22 +63,26 @@ void next_Click(object sender, System.EventArgs e)
...
-                Shell.GoNext();
+                // If it is not running, proceed to progress dialog
+                Shell.GoTo<Symphony.ProgressDialog>();
             }
         }
 
         void cancel_Click(object sender, System.EventArgs e)
         {
-            Shell.Cancel();
+            if( System.Windows.Forms.MessageBox.Show("Are you sure you want to cancel Symphony installation?",
```
VishwasShashidhar, line 81: same as above internationalisation comment

installer/win/WixSharpInstaller/CloseDialog.cs
mattias-symphony: I agree that it would be better, but we have not planned to do localization until phase 2 of installer, and figuring out and setting up to get this string from localization files would mean doing a big part of that work now, so I think that it is better to leave it for now. But I think I will add a "TODO: localization" comment to the relevant places in the code, so we can find it when we do localization

installer/win/WixSharpInstaller/ProgressDialog.cs
mattias-symphony: Well, it is not really commented out code - this file is a modified copy of a file from the Wix# source code, and these lines are documentation/examples, which might be useful for future work, so I think better to leave them in.

johankwarnmarksymphony (approved):

PR #1103, opened 2020-11-03
feat: SDA-2456: add screen annotate tool

VishwasShashidhar:
## Description
This PR adds support for the screen annotate component allowing users to draw, highlight & erase annotations.
[SDA-2456](https://perzoinc.atlassian.net/browse/SDA-2456)

## Solution Approach
- Create a new react component that is rendered in a new window
- Dynamically adjust the window size based on the image dimensions

## Related PRs
N/A

notes: add screen annotate tool


src/renderer/components/snipping-tool.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <button
+              className='ClearButton'
+              onClick={this.eventHandlers.onClear}
+            >
+              Clear
+            </button>
+          </div>
+        </header>
+
+        <main>
```
psjostrom, line 84: When presenting screen shots in the snippet tool, the screen shot is of different dimensions than what is being screen shoted. When the screen shot gets sent, it is in the original dimensions. This is actually a blocker for adding drawing capabilities, since the x and y coordinates are not the same in the tool and in the real screen shot, meaning when we add a SVG shape on and X and Y coordinate in the tool, those X and Y coordinates marks different things in the tool and in the real screen shot.


The screen shot presented in the tool should be of 1:1 dimensions as of what is being screen shoted in all cases, except if the screen shot is too big to fit the tool. In that case, the screen shot should be shrunk down (but of course keep the ratio). If the screen shot is being shown in different dimensions than its real dimensions, than the shapes will be off. Therefore, if the screen shot does not fit the tool, it must be shrunk down, not just change the presented dimensions, because of what I wrote before.


I think the wrong dimensions is only applied if the screen shot height > width. This can also be seen because  if heigth > width the screen shot gets a bit pixelated (tell tale sign that it is not beign shown in its correct dimensions). If width > heigth it is not.

src/renderer/components/snipping-tool.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Renders the Snipping tool
+   */
+  public render(): JSX.Element {
+    const { snipImage } = this.state;
+    return (
+      <div className='SnippingTool' lang={i18n.getLocale()}>
+        <header>
+          <div className='DrawActions'>
```
psjostrom, line 42: How do you set the border color of a button, I don't see it in the code but I may be missing it.

package.json
```diff
@@ -159,6 +159,7 @@
     "electron-spellchecker": "git+https://github.com/symphonyoss/electron-spellchecker.git#v2.0.4",
     "ffi-napi": "3.0.0",
     "filesize": "6.1.0",
+    "image-size": "^0.9.3",
```
psjostrom, line 162: Do you really need to use the image-size dependency? I sent a stack overflow to you earlier with a couple of different ways of geting it with vanilla javascript, amongst others there is just the imageElement.naturalHeight and naturalWidth properties on images. Coincidentally, that would be good to use to solve issue nr 1 as well, since naturalHeight and naturalWidth must be === presented-image-element-height and -width.

https://stackoverflow.com/questions/623172/how-to-get-image-size-height-width-using-javascript

psjostrom (changes requested):
When presenting screen shots in the snippet tool, the screen shot is of different dimensions than what is being screen shoted. When the screen shot gets sent, it is in the original dimensions. This is actually a blocker for adding drawing capabilities, since the x and y coordinates are not the same in the tool and in the real screen shot, meaning when we add a SVG shape on and X and Y coordinate in the tool, those X and Y coordinates marks different things in the tool and in the real screen shot.


The screen shot presented in the tool should be of 1:1 dimensions as of what is being screen shoted in all cases, except if the screen shot is too big to fit the tool. In that case, the screen shot should be shrunk down (but of course keep the ratio). If the screen shot is being shown in different dimensions than its real dimensions, than the shapes will be off. Therefore, if the screen shot does not fit the tool, it must be shrunk down, not just change the presented dimensions, because of what I wrote before.


I think the wrong dimensions is only applied if the screen shot height > width. This can also be seen because  if heigth > width the screen shot gets a bit pixelated (tell tale sign that it is not beign shown in its correct dimensions). If width > heigth it is not

package.json
psjostrom: Discussed in chat room, we cannot use these approaches because of NodeJS.

src/renderer/components/snipping-tool.tsx
psjostrom: Added now.

src/renderer/components/snipping-tool.tsx
psjostrom: Now img size is the same in tool and what is exported, but it is shrunk down in regards to what is actually being screen shoted. Creating a bug for this to track it.

psjostrom (approved):

src/app/screen-snippet-handler.ts
```diff
@@ -16,187 +23,279 @@ import { windowExists } from './window-utils';
...
+      width: dimensions.width,
+    };
+  }
+
+  /**
+   * Uploads a screen snippet
+   * @param webContents A browser window's web contents object
+   */
+  private uploadSnippet(webContents: Electron.webContents) {
+    ipcMain.on('upload-snippet', async (_event, snipImage: string) => {
```
KiranNiranjan, line 282: I see that this method is called every time `capture` is invoked. so, I think maybe we should we `ipcMain.once` to prevent memory leak?

KiranNiranjan (approved):
Apart from that change, it LGTM 👍

johankwarnmarksymphony (approved):

PR #1110, opened 2020-11-10
chore: SDA-2270: change clientVersion to sbeVersion

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-2270 About App: Change "clientVersion" to "sbeVersion" when writing the data to the clipboard

image

notes: change clientVersion to sbeVersion

### DIFF
--- a/src/renderer/components/about-app.tsx
+++ b/src/renderer/components/about-app.tsx
@@ -137,7 +137,6 @@ export default class AboutApp extends React.Component<{}, IState> {
             if (data) {
                 remote.clipboard.write({ text: JSON.stringify(data, null, 4) }, 'clipboard');
             }
-            remote.clipboard.write({ text: JSON.stringify(data2, null, 4) }, 'clipboard');
         }
     }
 
--- a/src/renderer/components/about-app.tsx
+++ b/src/renderer/components/about-app.tsx
@@ -130,13 +130,10 @@ export default class AboutApp extends React.Component<{}, IState> {
      * Copies the version info on to the clipboard
      */
     public copy(): void {
-        const data = this.state;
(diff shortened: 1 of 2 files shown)


src/renderer/components/about-app.tsx
```diff
@@ -133,7 +132,8 @@ export default class AboutApp extends React.Component<{}, IState> {
     public copy(): void {
         const data = this.state;
         if (data) {
-            remote.clipboard.write({ text: JSON.stringify(data, null, 4) }, 'clipboard');
+            const data2 = JSON.parse(JSON.stringify(data).split('"clientVersion":').join('"sbeVersion":'));
```
KiranNiranjan, line 135: What do you think about this?

```suggestion
            const { clientVersion, ...rest } = this.state;
            const data = { ...{ sbeVersion: clientVersion }, ...rest };
            if (data) {
                remote.clipboard.write({ text: JSON.stringify(data, null, 4) }, 'clipboard');
            }
```

KiranNiranjan (approved):

src/renderer/components/about-app.tsx
VishwasShashidhar: I recommend this as well.

VishwasShashidhar (approved):

VishwasShashidhar:
@johankwarnmarksymphony Looks like the unit tests are failing, you'll need to get a new snapshot from jest and push the changes.

For the above, just delete the snapshot `spec/__snapshots__/aboutApp.spec.ts.snap` and then run `npm run test`, that should automatically generate a new snapshot, commit that and push it.

mattias-symphony (approved):

PR #1127, opened 2020-11-30
fix: sda-2700 show 1.5/SFE 2.0/SFE-lite

johankwarnmarksymphony:
https://perzoinc.atlassian.net/browse/sda-2700 Showing SFE/SFE-lite version in About section

Notes: Showing SFE/SFE-lite version in About section

### DIFF
--- a/src/app/version-handler.ts
+++ b/src/app/version-handler.ts
@@ -1,6 +1,6 @@
 import { net } from 'electron';
 import * as nodeURL from 'url';
-import { buildNumber, clientVersion, optionalDependencies, searchAPIVersion, sfeVersion, version } from '../../package.json';
+import { buildNumber, clientVersion, optionalDependencies, searchAPIVersion, sfeClientType, sfeVersion, version } from '../../package.json';
 import { logger } from '../common/logger';
 import { config, IGlobalConfig } from './config-handler';
 
@@ -8,6 +8,7 @@ interface IVersionInfo {
     clientVersion: string;
     buildNumber: string;
     sfeVersion: string;
+    sfeClientType: string;
     sdaVersion: string;
     sdaBuildNumber: string;
     electronVersion: string;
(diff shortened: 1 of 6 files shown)


src/app/version-handler.ts
```diff
@@ -140,8 +142,22 @@ class VersionHandler {
 
             request.end();
 
+            logger.info('version-handler: mainUrl: ' + mainUrl);
+            logger.info('version-handler: hostname: ' + hostname);
+
             /* Get SFE version */
-            const urlSfeVersion = `${protocol}//${hostname}/client/version.json`;
+            let urlSfeVersion;
+            if (mainUrl?.includes('/client-bff/')) {
```
psjostrom, line 150: I don't think this works for local SFE-Lite clients, since they don't build urls in the same way as static. Kiran created an API for checking if you use mana or not on SFE-Lite, can you use that?

psjostrom (changes requested):

src/app/version-handler.ts
VishwasShashidhar: Agree with @psjostrom here. It also won't work if the deployment path changes from `client-bff` to something else in the future. I think we should use @KiranNiranjan's new API to find out if we are running Mana or not.

VishwasShashidhar (changes requested):
Apart from the way the client is checked, everything else looks good.

src/app/version-handler.ts
johankwarnmarksymphony: image

psjostrom: Haha, in that case it all comes down to if we can live with not getting the correct values when we run local.

VishwasShashidhar (approved):
Approving as per latest comment.

psjostrom (approved):

PR #1134, opened 2020-12-04
fix: Make devTools enabled dependent on if dev environment

psjostrom:
## Description
This PR makes devTools enabled or not for renderer processes dependent on environment.

Notes: engineering

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -896,9 +896,6 @@ export class WindowHandler {
         resizable: false,
         fullscreenable: false,
       },
-      {
-        devTools: isDevEnv,
-      },
     );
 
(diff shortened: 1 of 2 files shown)


mattias-symphony (approved):

KiranNiranjan:
@psjostrom we can simply override this change inside `getWindowOpts` right? instead of changing it in every place?

psjostrom:
> @psjostrom we can simply override this change inside `getWindowOpts` right? instead of changing it in every place?

Sure, calls for an API change for getWindowOpts, but I can of course do that also. Pushed commit with that change.

KiranNiranjan (approved):

PR #1145, opened 2020-12-10
fix: SDA-2762: load correct page on network issues

VishwasShashidhar:
## Description
For special cases like corp, load the actual pod url if login is not complete yet in case of network connection issues.
[SDA-2762](https://perzoinc.atlassian.net/browse/SDA-2762)

## Solution Approach
- Add a new flag called `isLoggedIn` in window handler
- Initialize the above flag to false by default
- Once SFE is loaded, set the flag to true

- If the above flag is false, load the original pod url
- If the flag is true, load the current window url

## Related PRs
N/A

Notes: load correct page on network issues

### DIFF
--- a/docs/features/network-error.md
+++ b/docs/features/network-error.md
@@ -1,18 +1,25 @@
 # Intro
-The SDA displays a network error page when there is some network related issue Including the error code. which makes it easier for users to understand what's the actual problem is.
+The SDA displays a network error page when there are network failures.
 
 # Platforms Supported
-macOS, Windows 10, Windows 7
+macOS, Windows 10
 
 # Purpose
-To let the users know the exact error the page stopped loading, instead of just displaying a blank screen.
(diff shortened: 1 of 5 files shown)


KiranNiranjan (approved):

docs/features/network-error.md
```diff
@@ -1,18 +1,25 @@
 # Intro
```
mattias-symphony, line 1: Great that documentation was updated as well

src/app/window-handler.ts
```diff
@@ -611,9 +612,7 @@ export class WindowHandler {
 
     // Reloads the Symphony
     ipcMain.on('reload-symphony', () => {
-      if (this.mainWindow && windowExists(this.mainWindow)) {
-        this.mainWindow.loadURL(this.url || this.globalConfig.url);
-      }
+      this.reloadSymphony();
```
mattias-symphony, line 615: Great that this is extracted to its own function

mattias-symphony (approved):

PR #1146, opened 2020-12-10
fix: SDA-2751 Fixed so BI can be sent from renderer process, etc

psjostrom:
## Description
Fixed so BI can be sent from renderer process, also fixed so annotate can be triggered with proper screen dimensions when alt tabed away. Also tidied up the logging from screen snippet.

[SDA-2571](https://perzoinc.atlassian.net/browse/SDA-2571)

Notes: Fixed BI, logging and alt tab with annotate

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -987,10 +987,17 @@ export class WindowHandler {
       width: number;
     },
   ): void {
-    const parentWindow = BrowserWindow.getFocusedWindow();
     // Prevents creating multiple instances
-    if (didVerifyAndRestoreWindow(this.snippingToolWindow) || !parentWindow) {
+    if (didVerifyAndRestoreWindow(this.snippingToolWindow)) {
       logger.error('window-handler: Could not open snipping tool window');
+      return;
+    }
+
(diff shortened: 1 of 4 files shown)


KiranNiranjan (approved):

src/app/window-handler.ts
```diff
@@ -987,21 +987,28 @@ export class WindowHandler {
...
-    const parentWindow = BrowserWindow.getFocusedWindow();
     // Prevents creating multiple instances
-    if (didVerifyAndRestoreWindow(this.snippingToolWindow) || !parentWindow) {
+    if (didVerifyAndRestoreWindow(this.snippingToolWindow)) {
       logger.error('window-handler: Could not open snipping tool window');
       return;
     }
 
+    const electronWindows = BrowserWindow.getAllWindows();
+    const mainWindow = electronWindows[0];
```
mattias-symphony, line 997: Will this work if BrowserWindow.getAllWindows() doesn't return any windows? Or maybe the assumption that there is always at least one window is fine

mattias-symphony (approved):

PR #1157, opened 2020-12-18
fix: SDA-2571 Fixed triggering multiple BI calls

psjostrom:
## Description
This PR fixes multiple BI calls being triggered

[SDA-2571](https://perzoinc.atlassian.net/browse/SDA-2571)

Notes: Skalman

### DIFF
--- a/src/app/screen-snippet-handler.ts
+++ b/src/app/screen-snippet-handler.ts
@@ -53,7 +53,7 @@ class ScreenSnippet {
       this.captureUtil = '/usr/bin/gnome-screenshot';
     }
 
-    ipcMain.on('send-tracking-data-to-main', async (_event, eventData: { element: AnalyticsElements, type: ScreenSnippetActionTypes }) => {
+    ipcMain.on('snippet-analytics-data', async (_event, eventData: { element: AnalyticsElements, type: ScreenSnippetActionTypes }) => {
       analytics.track({ element: eventData.element, action_type: eventData.type });
     });
   }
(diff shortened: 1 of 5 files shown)


KiranNiranjan (approved):

src/app/screen-snippet-handler.ts
```diff
@@ -52,6 +52,10 @@ class ScreenSnippet {
     if (isLinux) {
       this.captureUtil = '/usr/bin/gnome-screenshot';
     }
+
+    ipcMain.on('send-tracking-data-to-main', async (_event, eventData: { element: AnalyticsElements, type: ScreenSnippetActionTypes }) => {
```
VishwasShashidhar, line 56: Can we change the event name here? It looks very technical. It can be something like `snippet-event`

VishwasShashidhar (changes requested):
@psjostrom Please see comment.

src/app/analytics-handler.ts
```diff
@@ -66,6 +66,7 @@ class Analytics {
      * @param eventData {IAnalyticsData}
      */
     public track(eventData: IAnalyticsData): void {
+        console.warn('sending analytics data', eventData);
```
VishwasShashidhar, line 69: this is supposed to be a logger right? Not console log?

VishwasShashidhar (approved):

src/app/analytics-handler.ts
psjostrom: God darnit, not even a logger but was used during debugging....

PR #1172, opened 2021-01-14
feat: SDA-2824 Set notification position modal darkmode

psjostrom:
## Description

This PR adds the possibility to view notification position modal in dark mode, when dark mode is set on SFE-Lite.

### DIFF
--- a/src/app/main-api-handler.ts
+++ b/src/app/main-api-handler.ts
@@ -93,8 +93,9 @@ ipcMain.on(apiName.symphonyApi, async (event: Electron.IpcMainEvent, arg: IApiAr
             downloadHandler.setWindow(event.sender);
             break;
         case apiCmds.showNotificationSettings:
-            if (typeof arg.windowName === 'string' && typeof arg.theme === 'string') {
-                windowHandler.createNotificationSettingsWindow(arg.windowName, arg.theme);
+            if (typeof arg.windowName === 'string') {
+                const theme = arg.theme ? arg.theme : 'light';
+                windowHandler.createNotificationSettingsWindow(arg.windowName, theme);
             }
             break;
         case apiCmds.sanitize:
--- a/src/app/main-api-handler.ts
+++ b/src/app/main-api-handler.ts
@@ -93,8 +93,8 @@ ipcMain.on(apiName.symphonyApi, async (event: Electron.IpcMainEvent, arg: IApiAr
             downloadHandler.setWindow(event.sender);
             break;
         case apiCmds.showNotificationSettings:
-            if (typeof arg.windowName === 'string') {
-                windowHandler.createNotificationSettingsWindow(arg.windowName);
+            if (typeof arg.windowName === 'string' && typeof arg.theme === 'string') {
+                windowHandler.createNotificationSettingsWindow(arg.windowName, arg.theme);
             }
             break;
         case apiCmds.sanitize:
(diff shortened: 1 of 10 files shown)


src/app/main-api-handler.ts
```diff
@@ -93,8 +93,8 @@ ipcMain.on(apiName.symphonyApi, async (event: Electron.IpcMainEvent, arg: IApiAr
             downloadHandler.setWindow(event.sender);
             break;
         case apiCmds.showNotificationSettings:
-            if (typeof arg.windowName === 'string') {
-                windowHandler.createNotificationSettingsWindow(arg.windowName);
+            if (typeof arg.windowName === 'string' && typeof arg.theme === 'string') {
+                windowHandler.createNotificationSettingsWindow(arg.windowName, arg.theme);
```
KiranNiranjan, line 97: I think we should create a new API for updating theme settings on SDA instead of us send theme info in multiple APIs

psjostrom: Although I agree on one level, it would be more neat, in reality how SDA and SFE-Lite is coupled I don't agree. If we want to separate theme in its own API, we would have two alternatives.

1. Call for theme on loadup and make SFE-Lite send new theme on every update and save it on SDA side. This I would say is worst alternative - it would make for several sources of truth and we have had numerous bugs because of it.
2. Make an API on SFE-Lite and call it every time something needs to know current set theme. With this, we are doubling the calls between SDA and SFE-Lite (for every call from SFE-Lite to SDA to do X, SDA needs to request SFE-Lite and wait for response regarding theme).

If instead SFE-Lite sends theme when it tells SDA to do something that needs theme, we only having a single source of truth and are minimising calls between the two. We are essentially making a function call with the necessary information as parameters.

VishwasShashidhar: @psjostrom On point 1, It's more a case of data duplication than several sources of truth given we'll be updating the SDA side theme setting whenever there is a change on SFE-Lite. My only concern with point 2 (although it is a good solution) is that there'll be a back & forth every time you need to know the theme which may cause some latency.

src/renderer/ssf-api.ts
```diff
@@ -433,11 +433,12 @@ export class SSFApi {
     /**
      * Opens a modal window to configure notification preference.
      */
-    public showNotificationSettings(): void {
+    public showNotificationSettings(data: string): void {
```
VishwasShashidhar, line 436: @psjostrom I think we discussed about using an object here, any reason. you are using a string? And, the naming convention doesn't match the context either.

VishwasShashidhar (approved):
Looks good other than using a string in the API bridge @psjostrom

src/app/main-api-handler.ts
johankwarnmarksymphony: Theme-service on SFE-lite is the truth of source, I believe we should send the theme every time from SFE-lite to SDA.

Regarding reading theme from OS, that should be implemented as an update function to the truth of source theme-service. We can not use the OS method for setting dark mode in SDA. We should always get it from theme-service. If we don't do it like this, SDA can be dark and SFE-lite light.

johankwarnmarksymphony (approved):

src/app/main-api-handler.ts
KiranNiranjan, line 97: I think this would break in SFE 1.5 is that fine?

VishwasShashidhar: That's a good catch, it's true, this'll break 1.5. I think you should catch the theme argument as optional and set a default @psjostrom

PR #1188, opened 2021-02-09
SDA-2617: support windows native notifications

VishwasShashidhar:
## Description

This PR adds support for native notifications on Windows

**Notification**

Screenshot 2021-02-09 at 1 45 19 PM

**Notification Centre**

Screenshot 2021-02-09 at 1 46 32 PM

## Related PRs

N/A

### DIFF
--- a/src/app/init.ts
+++ b/src/app/init.ts
@@ -16,9 +16,11 @@ const userDataPath =
   userDataPathArg &&
   userDataPathArg.substring(userDataPathArg.indexOf('=') + 1);
 
-// force sandbox: true for all BrowserWindow instances.
+// If we are running in production, sandbox the entire app
+// and set the app user model id for windows native notifications
 if (!isNodeEnv) {
   app.enableSandbox();
+  app.setAppUserModelId('symphony_exe');
 }
 
 // need to set this explicitly if using Squirrel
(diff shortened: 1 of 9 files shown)


mattias-symphony (approved):

KiranNiranjan (approved):

src/app/init.ts
```diff
@@ -16,9 +16,11 @@ const userDataPath =
   userDataPathArg &&
   userDataPathArg.substring(userDataPathArg.indexOf('=') + 1);
 
-// force sandbox: true for all BrowserWindow instances.
+// If we are running in production, sandbox the entire app
+// and set the app user model id for windows native notifications
 if (!isNodeEnv) {
   app.enableSandbox();
+  app.setAppUserModelId('symphony_exe');
```
mattias-symphony, line 23: So is "symphony_exe" the same id as for the exe in the installer? I had assumed it was an internal id - if it is exposed, and used here, we might want to consider changing the id in the installer to something like "com.symphony....". I don't know :p

VishwasShashidhar: Yeah, that's correct. And, indeed, makes sense to put a proper name. But, for now, I'm sticking with this because I don't know the ramifications of changing it in the WixSharp installer :)

mattias-symphony: yes, agree

PR #1197, opened 2021-03-18
fix: SDA-3030: enable remote module explicitly

VishwasShashidhar:
## Description

Enable remote module explicitly

## Related PRs
#1198

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -1086,6 +1086,7 @@ export class WindowHandler {
       },
       {
         devTools: true,
+        enableRemoteModule: true,
       },
     );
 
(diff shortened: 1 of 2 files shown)


src/app/window-handler.ts
```diff
@@ -198,6 +198,7 @@ export class WindowHandler {
         },
         {
           preload: path.join(__dirname, '../renderer/_preload-main.js'),
+          enableRemoteModule: true,
```
KiranNiranjan, line 201: I think we can move this into `getWindowOpts` instead of updating it in all the places

KiranNiranjan (approved):

psjostrom (approved):

johankwarnmarksymphony (approved):

PR #1198, opened 2021-03-18
fix: SDA-3030: enable remote module explicitly

VishwasShashidhar:
Backport of #1197

### DIFF
--- a/scripts/build-win64.bat
+++ b/scripts/build-win64.bat
@@ -1,6 +1,6 @@
 :: Check to ensure that the VSDev command prompt is in the below location
 echo "Starting VSDev Command Prompt"
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\VsDevCmd.bat"
+call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Tools\VsDevCmd.bat"
 
 echo %PATH%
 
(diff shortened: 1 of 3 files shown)


src/app/window-handler.ts
```diff
@@ -196,6 +196,7 @@ export class WindowHandler {
         },
         {
           preload: path.join(__dirname, '../renderer/_preload-main.js'),
+          enableRemoteModule: true,
```
KiranNiranjan, line 199: I think we can move this into `getWindowOpts` instead of updating it in all the places

KiranNiranjan (approved):

scripts/build-win64.bat
```diff
@@ -1,6 +1,6 @@
 :: Check to ensure that the VSDev command prompt is in the below location
 echo "Starting VSDev Command Prompt"
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\VsDevCmd.bat"
+call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Tools\VsDevCmd.bat"
```
psjostrom, line 3: Wouldn't it be possible to check for both, so we support both versions of VS Studio?

johankwarnmarksymphony (approved):

scripts/build-win64.bat
VishwasShashidhar: Possible, but, not necessary since we can use 2017 to build older versions of SDA as well.

PR #1230, opened 2021-06-09
SDA-3210 Supporting boolean as msi arg

sbenmoussati:
## Description

Ability to set boolean values as args when installing SDA on Windows:

```
msiexec /i Symphony-Win64-9.3.0-909.msi OVERRIDE_USER_AGENT=true DEV_TOOLS_ENABLED=false
```

### DIFF
--- a/installer/win/WixSharpInstaller/Symphony.cs
+++ b/installer/win/WixSharpInstaller/Symphony.cs
@@ -373,18 +373,18 @@ public class CustomActions
             data = ReplaceProperty(data, "alwaysOnTop", session.Property("ALWAYS_ON_TOP"));
             data = ReplaceProperty(data, "launchOnStartup", session.Property("AUTO_START"));
             data = ReplaceProperty(data, "bringToFront", session.Property("BRING_TO_FRONT"));
-            data = ReplaceProperty(data, "media", session.Property("MEDIA"));
-            data = ReplaceProperty(data, "geolocation", session.Property("LOCATION"));
-            data = ReplaceProperty(data, "notifications", session.Property("NOTIFICATIONS"));
-            data = ReplaceProperty(data, "midiSysex", session.Property("MIDI_SYSEX"));
-            data = ReplaceProperty(data, "pointerLock", session.Property("POINTER_LOCK"));
-            data = ReplaceProperty(data, "fullscreen", session.Property("FULL_SCREEN"));
-            data = ReplaceProperty(data, "openExternal", session.Property("OPEN_EXTERNAL"));
             data = ReplaceProperty(data, "isCustomTitleBar", session.Property("CUSTOM_TITLE_BAR"));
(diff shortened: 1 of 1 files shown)


installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -408,10 +408,25 @@ static string ReplaceProperty( string data, string name, string value )
         // Using regular expressions to replace the existing value in the config file with the
         // one from the property. This is the same as the regex we used to have in the old
         // Advanced Installer, which looked like this: "url"\s*:\s*".*" => "url": "[POD_URL]"
-        return System.Text.RegularExpressions.Regex.Replace(data, @"""" + name + @"""\s*:\s*"".*""",
+        return System.Text.RegularExpressions.Regex.Replace(data, @"""" + name + @"""\s*:\s*"".[^,]*""",
```
mattias-symphony, line 411: Maybe I'm missing something, but why do we need this in the old ReplaceProperty?

installer/win/install_instructions_win.md
```diff
@@ -219,15 +219,15 @@ Expected values:
 
 #### Example, install with dev tools enabled
 
-    msiexec /i Symphony.msi DEV_TOOLS_ENABLED="true"
+    msiexec /i Symphony.msi DEV_TOOLS_ENABLED=true
```
mattias-symphony, line 222: I think that since the quotation marks on the command line gets stripped out, we should revert to specifying them in the documentation ( so "true" rather than true) just for consistency with other parameters like paths which might need them.  We've had some confusion in the path over what needs quotation marks and what doesn't, so even if not strictly necessary, I think it will make our support slightly easier to specify them for everything

mattias-symphony (approved):

installer/win/install_instructions_win.md
sbenmoussati: Totally agree with you. I’m reverting this part.

installer/win/WixSharpInstaller/Symphony.cs
sbenmoussati: I might be missing something as well, here are my observations.

The behaviour that I noticed is the following one for boolean's function:

first regex version ( @"""" + name + @"""\s*:\s*.*" ) was able to match the entire line, comma included: 
```"ALWAYS_ON_TOP": true,```

as a consequence, the replace method was changing the value of this property and was removing the comma:
```"ALWAYS_ON_TOP": false```. making JSON file not valid.

This is why I added  ```[^,]```` and did the same change for ```ReplaceProperty``` function.

mattias-symphony: Ah ok, I just thought it looked a bit strange with the comma inside the quotes, like it would match "name":"value," ?

sbenmoussati: Got your point. Indeed, it would catch the comma inside quotes. I'm reverting.

PR #1292, opened 2021-11-09
SDA improvement

sbenmoussati:
## Description
Faster dev env command and cleanup.
Avoiding reporting BI events for empty crashes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "main": "lib/src/app/init.js",
   "types": "lib/src/app/init.d.ts",
   "scripts": {
-    "postinstall": "npm run prebuild",
     "browserify": "run-s browserify-*",
     "browserify-preload": "browserify -o lib/src/renderer/_preload-main.js -x electron --insert-global-vars=__filename,__dirname lib/src/renderer/preload-main.js",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -168,6 +169,7 @@
     "ffi-napi": "3.0.0",
     "filesize": "6.1.0",
     "lazy-brush": "^1.0.1",
+    "lodash": "^4.17.21",
```
KiranNiranjan, line 172: I think we can just use `lodash.clonedeep`

src/app/main-api-handler.ts
```diff
@@ -413,3 +411,16 @@ ipcMain.handle(
     return;
   },
 );
+
+/**
+ * Log API call parameters.
+ */
+const logApiCallParams = (arg: any) => {
+  const apiCmd = arg.cmd;
+  const apiCmdParams = cloneDeep(arg);
```
KiranNiranjan, line 420: This is will very expensive right? I think we can just select what info we want to log?

KiranNiranjan (approved):
LGTM 👍🏻  apart from a couple of suggestions

PR #1331, opened 2022-01-26
SDA-3565 - Allow C2 / extensions to determine the presence of Citrix media redirection

rwestberg:
## Description

In order to support Citrix Media Optimization, it is necessary for an extension to be able to query the current status. Only when the redirection support is active is it necessary to proceed with loading and initializing the full Citrix client-side SDK.

### DIFF
--- a/src/app/citrix-handler.ts
+++ b/src/app/citrix-handler.ts
@@ -1,9 +1,29 @@
 import { enumerateValues, HKEY, RegistryValueType } from 'registry-js';
+import { isWindowsOS } from '../common/env';
 
-export const getCitrixMediaRedirectionStatus = ():
-  | 'inactive'
-  | 'supported'
-  | 'unsupported' => {
+export enum RedirectionStatus {
+  /**
+   * Citrix virtual environment is not active
+   */
+  INACTIVE = 'inactive',
+
+  /**
+   * Citrix virtual environment is active and media redirection is supported
+   */
+  SUPPORTED = 'supported',
+
+  /**
+   * Citrix virtual environment is active but media redirection is not supported
+   */
+  UNSUPPORTED = 'unsupported',
+}
+
+export const getCitrixMediaRedirectionStatus = (): RedirectionStatus => {
+  if (!isWindowsOS) {
+    // Citrix virtual environments are not supported on non-Windows OSes
+    return RedirectionStatus.INACTIVE;
+  }
+
   const values = enumerateValues(
     HKEY.HKEY_CURRENT_USER,
     'Software\\Citrix\\HDXMediaStream',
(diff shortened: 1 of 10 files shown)


src/app/citrix-handler.ts
```diff
@@ -0,0 +1,23 @@
+import { enumerateValues, HKEY, RegistryValueType } from 'registry-js';
+
+export const getCitrixMediaRedirectionStatus = ():
+  | 'inactive'
```
sbenmoussati, line 4: you can define here an enum that would have 'inactive' 'supported' or 'unsupported' values.

src/app/citrix-handler.ts
```diff
@@ -0,0 +1,23 @@
...
+  | 'supported'
+  | 'unsupported' => {
+  const values = enumerateValues(
+    HKEY.HKEY_CURRENT_USER,
+    'Software\\Citrix\\HDXMediaStream',
+  );
+
+  for (const value of values) {
+    if (value.name === 'MSTeamsRedirectionSupport') {
```
sbenmoussati, line 13: same here for 'MSTeamsRedirectionSupport'

src/renderer/ssf-api.ts
```diff
@@ -730,6 +730,18 @@ export class SSFApi {
   public supportedSettings(): string[] {
     return SUPPORTED_SETTINGS || [];
   }
+
+  /**
+   * Retrieves the current status of Citrix' media redirection feature
+   * @returns status
+   */
+  public getCitrixMediaRedirectionStatus(): Promise<
+    'inactive' | 'supported' | 'unsupported'
```
sbenmoussati, line 739: defined enum would be reused here

sbenmoussati:
globally LGTM, few comments added.
[Question]: this implementation is specific to Windows. Do we need to do something similar for macOS ?

KiranNiranjan (approved):
LGTM 👍🏻  and agree with the changes requested by @sbenmoussati

src/app/citrix-handler.ts
rwestberg: Makes sense, updated

rwestberg:
> globally LGTM, few comments added. [Question]: this implementation is specific to Windows, correct? Do we need to do something similar for macOS ?

I don't think there's anything needed for other OSes - this is only supported when running the client on a Citrix server which only exists on Windows (you can run the Citrix client on macOS to interact with it, but the SDA code would still be running on the Windows server).

sbenmoussati:
> > globally LGTM, few comments added. [Question]: this implementation is specific to Windows, correct? Do we need to do something similar for macOS ?
> 
> I don't think there's anything needed for other OSes - this is only supported when running the client on a Citrix server which only exists on Windows (you can run the Citrix client on macOS to interact with it, but the SDA code would still be running on the Windows server).

"Citrix server which only exists on Windows", that's the info I was missing, thanks for clarifying !

PR #1332, opened 2022-01-26
SDA-3561 - Allow C2 / extensions to retrieve the native window handle

rwestberg:
## Description

In order to support Citrix Media Optimization, the Citrix SDK requires that we can fetch the handle of the native window when playing video, in order to be able to place it correctly. This requires exposing the Electron function GetNativeWindowHandle to the SDA renderer process.

### DIFF
--- a/src/demo/win.html
+++ b/src/demo/win.html
@@ -1,87 +1,134 @@
 <html>
-<head>
-    <meta charset="UTF-8">
+  <head>
+    <meta charset="UTF-8" />
     <title>Test pop-out window</title>
-    <link rel="stylesheet" href="download-manager.css">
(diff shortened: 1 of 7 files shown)


src/demo/win.html
```diff
@@ -1,87 +1,134 @@
 <html>
-<head>
-    <meta charset="UTF-8">
+  <head>
```
rwestberg, line 2: The unrelated reformatting of this file is done by the pre-commit hook, not sure if I should try to avoid that..

sbenmoussati (approved):
👍

KiranNiranjan (approved):

PR #1337, opened 2022-02-02
SDA-3574 (Fix the condition to validate undefined)

KiranNiranjan:
## Description
- If the id value === 0 the `powerSaveBlocker` was never stopped
- Only Check if the id is undefined

### DIFF
--- a/src/app/app-state-handler.ts
+++ b/src/app/app-state-handler.ts
@@ -32,7 +32,7 @@ class AppStateHandler {
       this.id = powerSaveBlocker.start('prevent-display-sleep');
       logger.info('App-state: Prevent display to sleep, id: ' + this.id);
     } else {
-      if (this.id) {
+      if (this.id !== undefined) {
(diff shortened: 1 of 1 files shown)


src/app/app-state-handler.ts
```diff
@@ -32,7 +32,7 @@ class AppStateHandler {
       this.id = powerSaveBlocker.start('prevent-display-sleep');
       logger.info('App-state: Prevent display to sleep, id: ' + this.id);
     } else {
-      if (this.id) {
+      if (this.id !== undefined) {
```
sbenmoussati, line 35: so `blocker ID` can be 0 ... interesting :D

sbenmoussati (approved):

src/app/app-state-handler.ts
KiranNiranjan: Yeah, it starts from 0 😀

PR #1348, opened 2022-02-16
SDA-3532 Snipping tool should remain on the top on Windows and MacOS

sbenmoussati:
## Description
Snipping tool should remain on the top on Windows and macOS

### DIFF
--- a/src/app/window-handler.ts
+++ b/src/app/window-handler.ts
@@ -1250,7 +1250,7 @@ export class WindowHandler {
       opts.alwaysOnTop = true;
     }
 
-    if (isWindowsOS && this.mainWindow) {
+    if (this.mainWindow) {
       opts.parent = this.mainWindow;
     }
 


src/app/window-handler.ts
```diff
@@ -1250,7 +1250,7 @@ export class WindowHandler {
       opts.alwaysOnTop = true;
     }
 
-    if (isWindowsOS && this.mainWindow) {
+    if (this.mainWindow) {
       opts.parent = this.mainWindow;
```
KiranNiranjan, line 1254: On macOS this will attach the window to the main window amd ot won't be a separate window right?

KiranNiranjan (approved):

src/app/window-handler.ts
sbenmoussati: Here is a quick demo of its behaviour:

PR #1371, opened 2022-04-04
SDA-3626 Electron 17.3.0 upgrade

sbenmoussati:
## Description
Electron v17.3.0 upgrade

PR #1373, opened 2022-04-05
SDA-3626 Electron 17.3.0 upgrade (#1371)

sbenmoussati:
* SDA-3626 Electron 17.3.0 upgrade

* SDA-3626 Moving DesktopCapturer call to main process


package.json
```diff
@@ -17,7 +17,6 @@
     "browserify-preload-component": "browserify -o lib/src/renderer/_preload-component.js -x electron --insert-global-vars=__filename,__dirname lib/src/renderer/preload-component.js",
     "compile": "run-s lint compile:*",
     "compile:project": "gulp build",
-    "compile:spec": "tsc -p tsconfig.spec.json",
```
KiranNiranjan, line 20: We need this for unit tests right?

KiranNiranjan (approved):

package.json
sbenmoussati: As per my understanding, this tsconfig file is specific to Spectron which is not used anymore

PR #1410, opened 2022-05-05
SDA-3679 Ability to maximize SDA by dragging

sbenmoussati:
## Description
Bugfix: Ability to maximize SDA by dragging after putting the window in fullscreen.

### DIFF
--- a/src/app/window-actions.ts
+++ b/src/app/window-actions.ts
@@ -85,31 +85,17 @@ const saveWindowSettings = async (): Promise<void> => {
   }
 };
 
-const windowMaximized = async (): Promise<void> => {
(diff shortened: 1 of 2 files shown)


src/app/window-actions.ts
```diff
@@ -85,31 +85,17 @@ const saveWindowSettings = async (): Promise<void> => {
   }
 };
 
-const windowMaximized = async (): Promise<void> => {
```
sbenmoussati, line 88: as per my understanding this logic is already existing in `saveWindowSettings` function @KiranNiranjan

KiranNiranjan: Yes @sbenmoussati

KiranNiranjan (approved):

PR #1437, opened 2022-06-02
SDA-3733 (Upgrade electron version to 19)

KiranNiranjan:
## Description
Electron version: 19.0.1
NodeJS version: 16.13.2

- Upgrade electron version to 19
- Remove SwiftSearch dependency
- Upgrade ffi and ref modules
- Remove electron-rebuild dependency - use electron-builder's built preinstall script
- Clean up npm scripts


keerthi16 (approved):

src/app/main-api-handler.ts
```diff
@@ -48,7 +48,7 @@ import {
 let swiftSearchInstance;
 try {
   // tslint:disable-next-line:no-var-requires
-  const { SSAPIBridge } = require('swift-search');
+  const { SSAPIBridge } = {} as any; // require('swift-search');
```
sbenmoussati, line 51: [Question]: swift-search is not compatible with the upgrade to Electron 19 / Node 16?

KiranNiranjan: Yes, there are other major API changes that needs to be made in swift search module

sbenmoussati (approved):

PR #1447, opened 2022-06-28
 SDA-3589 (Auto update for mac and windows)

KiranNiranjan:
## Description
- Auto-update for mac and windows

## Demo
No updates available
Screenshot 2022-07-06 at 2 45 03 PM

Updates available
Screenshot 2022-07-06 at 2 47 22 PM

Update Downloaded
Screenshot 2022-07-06 at 2 49 04 PM

Download Complete
Screenshot 2022-07-06 at 2 51 37 PM

## Confluence
- https://perzoinc.atlassian.net/wiki/spaces/SDA/pages/2477754714/New+Auto+Update+electron+builder


keerthi16 (approved):

src/app/auto-update-handler.ts
```diff
@@ -1,67 +1,119 @@
...
     }
-  }
+
+    const url = userConfigURL ? userConfigURL : globalConfigURL;
+    const os = isMac ? MACOS : WINDOWS;
+
+    const { subdomain, domain, tld } = whitelistHandler.parseDomain(url);
+    logger.info(
+      `auto-update-handler: using generic pod url`,
+      `https://${subdomain}.${domain}.${tld}/${os}/${autoUpdateChannel}`,
```
sbenmoussati, line 112: It might be worth having a dedicated variable for the URL and reusing it as a return value.

src/app/auto-update-handler.ts
```diff
@@ -1,67 +1,119 @@
...
-      ];
+  public getUpdateUrl = (): string => {
+    const { url: userConfigURL } = config.getUserConfigFields(['url']);
+    const { url: globalConfigURL } = config.getGlobalConfigFields(['url']);
+    const { autoUpdateUrl, autoUpdateChannel } = config.getConfigFields([
+      'autoUpdateChannel',
+      'autoUpdateUrl',
+    ]);
+
+    if (autoUpdateUrl) {
```
sbenmoussati, line 98: Let's suppose autoUpdateUrl is a relative path: `/windows/general`
What would be the behaviour?

src/app/auto-update-handler.ts
```diff
@@ -1,67 +1,119 @@
...
+    const url = userConfigURL ? userConfigURL : globalConfigURL;
+    const os = isMac ? MACOS : WINDOWS;
+
+    const { subdomain, domain, tld } = whitelistHandler.parseDomain(url);
+    logger.info(
+      `auto-update-handler: using generic pod url`,
+      `https://${subdomain}.${domain}.${tld}/${os}/${autoUpdateChannel}`,
+    );
+
+    return `https://${subdomain}.${domain}.${tld}/${os}/${autoUpdateChannel}`;
```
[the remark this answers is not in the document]
sbenmoussati: Do we need to differentiate between the different OS? I believe yml file is already pointing to the right files.
If not, that would mean we can provide a default server URL in the user config which will work for both Windows and MacOS.

src/app/window-utils.ts
```diff
@@ -1077,6 +1088,35 @@ export const updateFeaturesForCloudConfig = async (
...
+        ]);
+        logger.info(
+          `window-utils: is last check date > auto update check interval?`,
+          lastAutoUpdateCheckDate,
+          autoUpdateCheckInterval,
+        );
+        // Compare the current date and user config last auto update checked date
+        // and if it is greater that autoUpdateCheckInterval we check for new updates
+        if (
+          getDifferenceInDays(new Date(), new Date(lastAutoUpdateCheckDate)) >
```
sbenmoussati, line 1109: `lastAutoUpdateCheckDate ` not being set the first time,  `new Date(lastAutoUpdateCheckDate)`. will fail.
I think it's better to check if we have a valid value when we get it from the config file. If it's not a valid one (empty string, undefined) then we can set the value in config file and run `autoUpdate.checkUpdates()` , wdyt?

config/Symphony.config
```diff
@@ -1,5 +1,9 @@
 {
     "url":"https://my.symphony.com",
+    "autoUpdateUrl": "https://27e5-165-225-121-53.in.ngrok.io/Test/",
```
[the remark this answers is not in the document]
KiranNiranjan: Yes

src/app/auto-update-handler.ts
```diff
@@ -1,67 +1,119 @@
...
+          mainWebContents.send('display-client-banner', {
+            reason: 'autoUpdate',
+            action: 'update-not-available',
+          });
+        }
+      });
 
-    this.updateUtilArgs = [];
+      this.autoUpdater.on('update-downloaded', () => {
+        this.isUpdateAvailable = true;
```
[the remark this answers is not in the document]
KiranNiranjan: Already attached `electron-log` to `auto updater` so, anything related to it will be automatically logged

src/app/auto-update-handler.ts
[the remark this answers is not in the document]
KiranNiranjan: Yes, I have updated the artifacts with os type so, that `auto updater` will pick the correct .zip file

src/app/auto-update-handler.ts
KiranNiranjan: auto updater will fail to fetch the updates and logs error. Should we have any validation in this case and direct it to using the channel instead?

sbenmoussati: I think that option would be preferable yes!

sbenmoussati (approved):
👍 nice!! One small comment on a UT name.

keerthi16 (approved):

PR #1454, opened 2022-07-12
SDA-3589 (Remove unwanted targets and add extra files)

KiranNiranjan:
## Description
- Add extra files (screen snippet, cloud9 & screen share indicator frame)
- Remove nsis and dmg build targets

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "asar": true,
     "asarUnpack": "**/*.node",
     "nsis": {
-      "perMachine": true,
+      "perMachine": false,
       "oneClick": false,
       "allowToChangeInstallationDirectory": true,
       "include": "build/installer.nsh",
--- a/package.json
+++ b/package.json
@@ -41,14 +41,14 @@
   },
   "build": {
     "appId": "com.symphony.electron_desktop",
-    "uninstallDisplayName": "${productName}",
     "artifactName": "${productName}-${version}-${os}.${ext}",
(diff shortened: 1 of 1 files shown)


keerthi16 (approved):

package.json
```diff
@@ -41,14 +41,15 @@
   },
   "build": {
     "appId": "com.symphony.electron_desktop",
-    "uninstallDisplayName": "${productName}",
     "artifactName": "${productName}-${version}-${os}.${ext}",
     "asar": true,
     "asarUnpack": "**/*.node",
     "nsis": {
```
sbenmoussati, line 47: [Question]: we still have an nsis build entry. However, nsis target is being removed. Is there a specific reason for that?

package.json
```diff
@@ -41,14 +41,15 @@
...
       "perMachine": true,
```
sbenmoussati, line 48: [Question] does it mean we would need admin rights to perform an auto update?

package.json
KiranNiranjan: @sbenmoussati instead of removing `zip` I had removed nsis here.

package.json
KiranNiranjan: Admin permission is required only if the user is installing in the program files directory

`perMachine` and `oneClick` will be **false** and the user will be shown an option to choose the install directory

sbenmoussati (approved):

PR #1467, opened 2022-08-04
SDA-3780 (Read user config before starting the application)

KiranNiranjan:
## Description
- Rever the app id changes 
- Read user config before starting the application

### DIFF
--- a/installer/win/WixSharpInstaller/Symphony.cs
+++ b/installer/win/WixSharpInstaller/Symphony.cs
@@ -35,7 +35,7 @@ class Script
         // desired contents of installation, and then we can simplify this bit.
         var project = new ManagedProject(productName,
             new Dir(@"%ProgramFiles%\" + productName,
-                new File(new Id("com.symphony.electron-desktop"), @"..\..\..\dist\win-unpacked\Symphony.exe",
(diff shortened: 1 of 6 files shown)


keerthi16 (approved):

installer/win/WixSharpInstaller/Symphony.cs
```diff
@@ -35,7 +35,7 @@ static public void Main(string[] args)
         // desired contents of installation, and then we can simplify this bit.
         var project = new ManagedProject(productName,
             new Dir(@"%ProgramFiles%\" + productName,
-                new File(new Id("com.symphony.electron_desktop"), @"..\..\..\dist\win-unpacked\Symphony.exe",
+                new File(new Id("symphony_exe"), @"..\..\..\dist\win-unpacked\Symphony.exe",
```
sbenmoussati, line 38: Do we still need to revert app Id?

KiranNiranjan: Not required, Should I update to the new one?

sbenmoussati: if it has no impact on auto-update then no need to update it to the new one.

sbenmoussati (approved):

PR #1473, opened 2022-08-18
SDA-3815 (Disable maximize button and enable minimize button)

KiranNiranjan:
## Description
- Disable only maximize button and not minimize button to match the SDA 9.2 version

#### Custom Title bar with maximize button disabled
Screenshot 2022-08-18 at 3 20 07 PM

#### Native Title bar with maximize button disabled
Screenshot 2022-08-18 at 3 21 58 PM

### DIFF
--- a/src/renderer/components/windows-title-bar.tsx
+++ b/src/renderer/components/windows-title-bar.tsx
@@ -81,7 +81,7 @@ export default class WindowsTitleBar extends React.Component<{}, IState> {
    * Renders the component
    */
   public render(): JSX.Element | null {
-    const { title, isDisabled } = this.state;
+    const { title } = this.state;
 
     return (
       <div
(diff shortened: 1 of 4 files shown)


keerthi16 (approved):

src/renderer/components/windows-title-bar.tsx
```diff
@@ -128,7 +128,6 @@ export default class WindowsTitleBar extends React.Component<{}, IState> {
             onClick={this.eventHandlers.onMinimize}
             onContextMenu={this.eventHandlers.onDisableContextMenu}
             onMouseDown={this.handleMouseDown}
-            disabled={isDisabled}
           >
             <svg x='0px' y='0px' viewBox='0 0 14 1'>
               <rect fill='rgba(255, 255, 255, 0.9)' width='14' height='0.6' />
```
sbenmoussati, line 133: IMO it would be preferable to change svg color (maybe something like rgba(149, 149, 149, 0.9) rather than background color + disable button hovering

sbenmoussati (approved):

PR #1482, opened 2022-08-30
SDA-3839 Resize SDA if unmaximizing

sbenmoussati:
## Description
`unmaximize` event is no more triggered on Electron v19. This regression has been introduced in Electron v18.4 (cannot reproduce on v18.3). As a consequence, `on('unmaximize')` function isn't triggered anymore.
As a work-around, we force app resizing if we need to unmaximize SDA.

### DIFF
--- a/src/app/main-api-handler.ts
+++ b/src/app/main-api-handler.ts
@@ -333,9 +333,14 @@ ipcMain.on(
       case apiCmds.unmaximizeMainWindow:
         const mainWindow = windowHandler.getMainWindow();
         if (mainWindow && windowExists(mainWindow)) {
-          mainWindow.isFullScreen()
-            ? mainWindow.setFullScreen(false)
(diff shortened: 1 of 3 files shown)


src/app/main-api-handler.ts
```diff
@@ -333,9 +333,14 @@ ipcMain.on(
...
         if (mainWindow && windowExists(mainWindow)) {
-          mainWindow.isFullScreen()
-            ? mainWindow.setFullScreen(false)
-            : mainWindow.unmaximize();
+          if (mainWindow.isFullScreen()) {
+            mainWindow.setFullScreen(false);
+          } else {
+            mainWindow.unmaximize();
+            setTimeout(() => {
+              windowHandler.forceUnmaximize();
```
KiranNiranjan, line 341: Nice

KiranNiranjan (approved):

PR #1511, opened 2022-09-29
SDA-3892: About Symphony redesign

NguyenTranHoangSym:
## Description
- MacOS - Change "About Symphony" window into a modal, user will not be able to click on back window while about window is still on
- Add "OK" button to the modal to close the modal


src/app/window-handler.ts
```diff
@@ -1119,7 +1119,7 @@ export class WindowHandler {
       opts.alwaysOnTop = true;
     }
 
-    if (isWindowsOS && selectedParentWindow) {
+    if ((isMac || isWindowsOS) && selectedParentWindow) {
```
sbenmoussati, line 1122: as we just support Windows and macOS, you can just remove OS conditions
```suggestion
    if (selectedParentWindow) {
```

src/locale/en.json
```diff
@@ -7,7 +7,8 @@
     "Others": "Others",
     "Swift Search": "Swift Search",
     "Swift Search API": "Swift Search API",
-    "Symphony Logo": "Symphony Logo"
+    "Symphony Logo": "Symphony Logo",
+    "Ok": "Ok"
```
sbenmoussati, line 11: Crowdin is not configured on SDA, we also need to add the different entries in Fr and JP i18n files

src/renderer/styles/about-app.less
```diff
@@ -121,7 +121,30 @@ body {
     font-size: 0.8rem;
     text-align: center;
     padding: 10px 32px;
-    margin: 4px 20px;
+    margin: 4px 8px;
+    display: inline-block;
+    text-decoration: none;
+    line-height: 12px;
+    background-color: rgba(61, 162, 253, 1);
```
sbenmoussati, line 128: You can define a variable for this color

src/renderer/styles/about-app.less
```diff
@@ -121,7 +121,30 @@ body {
...
+    color: rgba(255, 255, 255, 1);
```
sbenmoussati, line 129:
```suggestion
    color: @white;
```

src/renderer/components/about-app.tsx
```diff
@@ -175,6 +184,13 @@ export default class AboutApp extends React.Component<{}, IState> {
     }
   }
 
+  /**
+   * Confirm to close modal
+   */
+  public confirm(): void {
+    window.close();
```
KiranNiranjan, line 191: I think we need to close the window from the main process

sbenmoussati: Agree, @NguyenTranHoangSym have a look at the way "copy" function is implemented. This API allows IPC

src/locale/en.json
NguyenTranHoangSym: Possibly, but lets ask for the translation from Crowdin so I can manually add it. I have also added few more keys, where should I send these keys to get the translation?

src/renderer/styles/about-app.less
NguyenTranHoangSym: Im going to change these into a variable that we can possibly adapt with the Design System of Symphony
I.E: [Electricity05] -> 
image

src/renderer/components/about-app.tsx
NguyenTranHoangSym: the dialog itself has its own window life cycle. I dont seem to find a method trigger destroying a modal from BrowserWindow. What I can possibly do is to listen to the destroy event and do certain thing, like gain back focus

NguyenTranHoangSym: I see, let me see how should it be done, its not usual to pass callback.

sbenmoussati: Basically:
- we send an event to the main process on a dedicated channel (ipcRenderer.send)
- main process is listening to a channel (ipcMain)
- once we receive the event, we perform the action (here, we close the window and destroy its reference in window-handler.ts)

src/app/window-handler.ts
NguyenTranHoangSym: Done!

src/locale/en.json
NguyenTranHoangSym: Awaiting for new translation @sbenmoussati . Shoud I ping in Crowdin room?

src/renderer/styles/about-app.less
NguyenTranHoangSym: Done, adding new color palette section

src/renderer/styles/about-app.less
NguyenTranHoangSym: this has been removed since it belongs to old design I believe

src/renderer/components/about-app.tsx
```diff
@@ -175,6 +208,13 @@ export default class AboutApp extends React.Component<{}, IState> {
     }
   }
 
+  /**
+   * Confirm to close modal
+   */
+  public confirm(): void {
```
sbenmoussati, line 214: I think `close` would be more appropriate here

sbenmoussati (commented):
Looks good to me 👍 , added small comments

src/renderer/components/about-app.tsx
```diff
@@ -89,58 +90,90 @@ export default class AboutApp extends React.Component<{}, IState> {
...
+                </li>
+              ))}
             </ul>
           </section>
+          <div className='AboutApp-copy-container'>
+            <button
+              className='AboutApp-copy-button'
+              onClick={this.eventHandlers.onCopy}
+              title={i18n.t(
+                'Copy all the version information in this page',
```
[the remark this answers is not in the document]
NguyenTranHoangSym: Done, I have changed everything, if you can recheck @sbenmoussati

KiranNiranjan (approved):

crowdin.yml
```diff
@@ -1,29 +1,28 @@
...
-      "translation": "/src/locale/%locale%.json",
+      'source': '/src/locale/en-US.json',
+      'translation': '/src/locale/%locale%.json',
     },
     # Duplicate file
     {
-      "source": "/src/locale/en.json",
-      "translation": "/src/locale/%two_letters_code%.json",
+      'source': '/src/locale/en.json',
+      'translation': '/src/locale/%two_letters_code%.json',
```
sbenmoussati, line 26: Question: what is the purpose of %two_letters_code%?

sbenmoussati (approved):

crowdin.yml
```diff
@@ -1,29 +1,28 @@
 #
 # Your Crowdin credentials
 #
-"project_id": "420172"                    # Crowdin project Settings > API & Webhooks > API > Project Id
-"api_token_env": "CROWDIN_PERSONAL_TOKEN" # Low priority: superseded by user-specific credentials file ~/.crowdin.yaml
-"base_path": "."
-"base_url": "https://api.crowdin.com"
+'project_id': '415116' # Crowdin project Settings > API & Webhooks > API > Project Id
```
sbenmoussati, line 4: Let me double check for the project ID

crowdin.yml
NguyenTranHoangSym: two letter codes as I understand they will take 2 letters short for the language.
I.E: vn
or %locale% to be en-US

NguyenTranHoangSym: @sbenmoussati 
for more information
image

crowdin.yml
NguyenTranHoangSym: Allright!!

sbenmoussati: We can keep using 420172 @NguyenTranHoangSym

NguyenTranHoangSym: So that was the original. Can you also take a look if we have any keys tranlsated already there

sbenmoussati (approved):

PR #1519, opened 2022-10-18
SDA-3901 (Add new implementation for welcome screen)

KiranNiranjan:
## Description
- Add new implementation for the welcome screen

Screenshot 2022-10-18 at 12 54 38 PM

Screenshot 2022-10-18 at 12 52 27 PM


keerthi16 (approved):

src/app/window-handler.ts
```diff
@@ -233,6 +232,13 @@ export class WindowHandler {
     this.isCustomTitleBar =
       isWindowsOS &&
       this.config.isCustomTitleBar === CloudConfigDataTypes.ENABLED;
+    // Get url to load from cmd line or from global config file
+    const urlFromCmd = getCommandLineArgs(process.argv, '--url=', false);
+    this.isPodConfigured = !config.isFirstTimeLaunch();
+    this.didShowWelcomeScreen = false;
+    this.shouldShowWelcomeScreen =
```
sbenmoussati, line 239: I gave it a try on st3 (sso not being configured) and I believe we should also check if the pod has sso. If SSO is not configured, we should proceed with normal flow and add log something about it.

About SSO:
- we can get SSO status by relying on `/login/checkauth?type=user` endpoint
- this endpoint response contains a field `authenticationType`. If `authenticationType` === `password` then we can skip opening the browser

sbenmoussati (approved):
Impressive! 👏 
Small improvement regarding SSO and browser based authentication check

sbenmoussati (approved):
👏🏻

baphony:
Hi @KiranNiranjan,
FYI, I am getting this error (on the last commit df14fa1), which I don't have on the prior commit (fb5a981):
image

PR #1532, opened 2022-11-04
RTC-13524 Add notification ignore button

baphony:
## Description

For Symphony-C9 extension we need to be able to ignore an incoming call from a notification.
These changes will display an `Ignore` button on notifications when a new `hasIgnore: true` notification parameter is added.
A click on this button will trigger a `notification-ignore` event.
This is very similar to the existing notification `Reply` button.

### DIFF
--- a/spec/notificationComp.spec.ts
+++ b/spec/notificationComp.spec.ts
@@ -129,6 +129,19 @@ describe('Toast notification component', () => {
     expect(toastNotificationReplyButton.exists()).toBeTruthy();
   });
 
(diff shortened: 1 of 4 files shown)


baphony:
I created another PR targeting branch 19.0.x (as there is a bug which doesn't allow to use the --url param on the main branch, I used the branch 19 to test it):
https://github.com/finos/SymphonyElectron/pull/1531

sbenmoussati (approved):
LGTM 👍

PR #1545, opened 2022-11-16
SDA-3966 Support Paste as plain text

sbenmoussati:
## Description
Support `Paste as plain text`

### DIFF
--- a/src/app/context-menu-builder.ts
+++ b/src/app/context-menu-builder.ts
@@ -14,6 +14,7 @@ interface IContextMenuStringTable {
   cut: () => string;
   copy: () => string;
   paste: () => string;
+  pasteAsPlainText: () => string;
   inspectElement: () => string;
 }
 
@@ -29,6 +30,7 @@ const contextMenuStringTable: IContextMenuStringTable = {
   cut: () => `Cut`,
   copy: () => `Copy`,
   paste: () => `Paste`,
+  pasteAsPlainText: () => `Paste as plain text`,
   inspectElement: () => `Inspect Element`,
 };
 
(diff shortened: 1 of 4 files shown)


src/app/context-menu-builder.ts
```diff
@@ -433,6 +435,21 @@ export class ContextMenuBuilder {
     return menu;
   }
 
+  /**
+   * Adds the Paste menu item.
+   */
+  public addPasteAndMatchStyle(menu, menuInfo) {
```
NguyenTranHoangSym, line 441: should this be renamed into addPasteAsPlainText? since the meaning of whole thing is for doing this

sbenmoussati: depends if we're considering chromium definition or functional def.

src/app/context-menu-builder.ts
```diff
@@ -433,6 +435,21 @@ export class ContextMenuBuilder {
...
+  /**
+   * Adds the Paste menu item.
+   */
+  public addPasteAsPlainText(menu, menuInfo) {
+    menu.append(
+      new MenuItem({
+        label: this.stringTable.pasteAsPlainText(),
+        accelerator: 'CommandOrControl+Shift+V',
+        enabled: menuInfo.editFlags.canPaste,
+        role: 'pasteAndMatchStyle',
```
NguyenTranHoangSym, line 447: is this suppose to be pasteAsPlainText?

sbenmoussati: yes, which correspond to role "pasteAndMatchStyle" on chromium

NguyenTranHoangSym (approved):

PR #1579, opened 2022-12-01
SDA-3996 Bring to front on notification bugfix

sbenmoussati:
## Description
Backporting #1578

### DIFF
--- a/src/app/window-actions.ts
+++ b/src/app/window-actions.ts
@@ -115,8 +115,10 @@ const throttledWindowChanges = throttle(async (eventName, window) => {
 const throttledWindowRestore = throttle(async () => {
   notification.moveNotificationToTop();
   const mainWebContents = windowHandler.getMainWebContents();
-  if (mainWebContents && !mainWebContents.isDestroyed()) {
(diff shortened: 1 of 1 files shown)


src/app/window-actions.ts
```diff
@@ -115,8 +115,10 @@ const throttledWindowChanges = throttle(async (eventName, window) => {
 const throttledWindowRestore = throttle(async () => {
   notification.moveNotificationToTop();
   const mainWebContents = windowHandler.getMainWebContents();
-  if (mainWebContents && !mainWebContents.isDestroyed()) {
-    mainWebContents.focus();
+  if (isWindowsOS) {
```
NguyenTranHoangSym, line 118: so this will only apply for windows?

sbenmoussati: mainWebContents.focus() is buggy on macOS and putting SDA to the background

NguyenTranHoangSym: I see approved

NguyenTranHoangSym (approved):

PR #1661, opened 2022-12-28
SDA-4036 Remove 1.5 related code for version discovery

sbenmoussati:
## Description
Version field should always display 2.0 version in "About" window.

### DIFF
--- a/src/renderer/components/about-app.tsx
+++ b/src/renderer/components/about-app.tsx
@@ -35,6 +35,7 @@ interface IState {
 }
 
 const ABOUT_SYMPHONY_NAMESPACE = 'AboutSymphony';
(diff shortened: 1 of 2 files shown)


src/renderer/components/about-app.tsx
```diff
@@ -118,9 +117,7 @@ export default class AboutApp extends React.Component<{}, IState> {
       },
     ];
 
-    if (sfeClientType !== '1.5') {
-      sfeClientTypeName = 'SFE-Lite';
-    }
+    sfeClientTypeName = 'SFE-Lite';
```
NguyenTranHoangSym, line 120: Line 101 has this already. can we just remove this?

NguyenTranHoangSym (approved):

PR #1682, opened 2023-01-10
SDA-4035 (Prevent auto-focus on copy button on component loading)

sbenmoussati:
## Description
This is a side-effect of FOUC optimization:
Post FOUC optim, created window is displayed by calling `BrowserWindow.show()` method which automatically bring focus on current window. On macOS, this will automatically bring focus on "Copy" button. Goal is to blur it.

### DIFF
--- a/src/renderer/components/about-app.tsx
+++ b/src/renderer/components/about-app.tsx
@@ -44,9 +44,11 @@ export default class AboutApp extends React.Component<{}, IState> {
     onCopy: () => this.copy(),
     onClose: () => this.close(),
   };
+  private copyButtonRef: React.RefObject<HTMLButtonElement>;
 
   constructor(props) {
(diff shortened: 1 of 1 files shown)


src/renderer/components/about-app.tsx
```diff
@@ -188,6 +191,9 @@ export default class AboutApp extends React.Component<{}, IState> {
    * Callback to handle event when a component is mounted
    */
   public componentDidMount(): void {
+    setTimeout(() => {
+      this.copyButtonRef.current?.blur();
```
NguyenTranHoangSym, line 195: Same, this can be merged but Windows is always unfaithful to timing like this? Just a Question @sbenmoussati

NguyenTranHoangSym (approved):

PR #1688, opened 2023-01-13
SDA-4035 Blue border while focus is on copy button

sbenmoussati:
## Description
Blue border while focus is on copy button

### DIFF
--- a/src/renderer/components/about-app.tsx
+++ b/src/renderer/components/about-app.tsx
@@ -45,11 +45,11 @@ export default class AboutApp extends React.Component<{}, IState> {
     onCopy: () => this.copy(),
     onClose: () => this.close(),
   };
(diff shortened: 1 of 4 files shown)


src/renderer/components/about-app.tsx
```diff
@@ -134,16 +134,7 @@ export default class AboutApp extends React.Component<{}, IState> {
...
-            <ul className='AboutApp-symphony-section'>
-              {symphonySectionItems.map((item, key) => (
-                <li key={key}>
-                  <strong>{item.key}</strong>
-                  <span>{item.value}</span>
-                </li>
-              ))}
-            </ul>
-          </section>
+          {sdaVersion ? this.renderVersions(symphonySectionItems) : null}
```
NguyenTranHoangSym, line 137: We can just skip the null value, if it has value then this render things else nothing happen

PR #1689, opened 2023-01-13
SDA-4049 Taking screenshot - hiding fullscreen window

sbenmoussati:

### DIFF
--- a/src/app/stores/window-store.ts
+++ b/src/app/stores/window-store.ts
@@ -1,5 +1,6 @@
 import { BrowserWindow } from 'electron';
-import { isMac } from '../../common/env';
+import { isMac, isWindowsOS } from '../../common/env';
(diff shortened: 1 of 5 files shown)


src/app/stores/window-store.ts
```diff
@@ -50,53 +51,50 @@ export class WindowStore {
...
+        );
+        currentWindow!.focused = true;
       }
-    }
-  };
-
-  public restoreWindowsOnCapturing = (hideOnCapture?: boolean) => {
-    if (hideOnCapture) {
-      const currentWindows = this.getWindowStore();
+      const focusedWindowDetails = currentWindows.windows.find(
```
NguyenTranHoangSym, line 67: shouldn't we use the currentWindow above?

PR #1708, opened 2023-02-16
RTC-13831 Stop C9Shell process when entering sleep/hibernation mode

baphony:
This will improve battery consumption and fix C9 issues (can't call, hang up, change DND mode...)

https://perzoinc.atlassian.net/browse/RTC-13831

### DIFF
--- a/src/app/c9-shell-handler.ts
+++ b/src/app/c9-shell-handler.ts
@@ -1,4 +1,4 @@
-import { app, WebContents } from 'electron';
+import { app, powerMonitor, WebContents } from 'electron';
 import { isDevEnv, isWindowsOS } from '../common/env';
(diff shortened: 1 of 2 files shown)


src/app/c9-shell-handler.ts
```diff
@@ -21,6 +21,16 @@ class C9ShellHandler {
   private _curStatus: IShellStatus | undefined;
   private _statusCallback: StatusCallback | undefined;
 
+  constructor() {
```
swasunb, line 24: add some unit tests may be ?

baphony: I didn't manage to mock/spy on powerMonitor.on, I have spent too much on this, I give up 🙃

PR #1710, opened 2023-02-20
SDA-3900 Cookies alignment with PRD

sbenmoussati:
## Description
Cookies alignment with PRD
Electron upgrade
Adding missing shortcuts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "symphony",
-  "version": "23.1.0",
+  "version": "23.4.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "symphony",
-      "version": "23.1.0",
+      "version": "23.4.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


src/app/protocol-handler.ts
```diff
@@ -112,29 +113,46 @@ class ProtocolHandler {
...
-      if (anticsrfValue) {
-        await session.defaultSession.cookies.set({
-          url,
-          name: 'anti-csrf-cookie',
-          value: anticsrfValue,
-        });
+      if (skeyValue && anticsrfValue) {
+        try {
+          await session.defaultSession.cookies.set({
+            url,
```
NguyenTranHoangSym, line 128: can we define an interface for this object for reusing purpose and for indication?

src/app/protocol-handler.ts
```diff
@@ -112,29 +113,46 @@ class ProtocolHandler {
...
+            name: 'skey',
+            value: skeyValue,
+            secure: true,
+            httpOnly: true,
+            sameSite: 'no_restriction',
+            domain: cookieDomain,
+          });
+
+          await session.defaultSession.cookies.set({
+            url,
```
NguyenTranHoangSym, line 138: same here, should have 1 interface for both? I understand you are setting 2 different type of cookies but there are some common sense between them that we can group

NguyenTranHoangSym (approved):
LGTM

PR #1711, opened 2023-02-20
SDA-4080 Adding support to startpage for C2 switch between stable and daily

sbenmoussati:
## Description
Adding support to startpage for C2 switch between stable and daily

### DIFF
--- a/src/app/app-menu.ts
+++ b/src/app/app-menu.ts
@@ -619,8 +619,6 @@ export class AppMenu {
       (windowHandler.url &&
         windowHandler.url.startsWith('https://corporate.symphony.com')) ||
       false;
-    const isBFFServedContent =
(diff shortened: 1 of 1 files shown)


src/app/app-menu.ts
```diff
@@ -715,22 +716,32 @@ export class AppMenu {
           id: C2_CHANNELS_MENU_ID,
           submenu: [
             {
-              click: (_item) =>
-                windowHandler.switchClient(ClientSwitchType.CLIENT_2_0),
+              click: (_item) => {
```
NguyenTranHoangSym, line 719: better create a dedicated function with arrow function declaration, this would help in performance issue.

NguyenTranHoangSym (approved):

PR #1921, opened 2023-07-31
SDA-4234: Support Tel Protocol on loading _blank

NguyenTranHoangSym:
## Description
SDA will now support +tel on opening a new page with target=_blank


src/app/window-utils.ts
```diff
@@ -130,7 +131,7 @@ export const preventWindowNavigation = (
   );
 
   const listener = async (e: Electron.Event, winUrl: string) => {
-    if (!winUrl.startsWith('https')) {
+    if (!verifyProtocolForNewUrl(winUrl)) {
```
KiranNiranjan, line 134: Do we need this here?

NguyenTranHoangSym: done I have removed it

KiranNiranjan (approved):

PR #1939, opened 2023-08-16
SDA-4270 config file persistency bugfix

sbenmoussati:
## Description
Config file persistency bugfix

### DIFF
--- a/src/app/config-handler.ts
+++ b/src/app/config-handler.ts
@@ -147,6 +147,7 @@ class Config {
   public filteredCloudConfig: ICloudConfig | {};
   private isFirstTime: boolean = true;
   private didUpdateConfigFile: boolean = false;
+  private isUpdatingConfigFile: boolean = false;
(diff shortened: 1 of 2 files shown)


src/app/config-handler.ts
```diff
@@ -222,18 +223,25 @@ class Config {
...
       if (!this.didUpdateConfigFile) {
+        this.isUpdatingConfigFile = true;
         event.preventDefault();
-        const id = powerSaveBlocker.start('prevent-app-suspension');
         logger.info(
           `config-handler: power save blocker id ${id} and is started`,
           powerSaveBlocker.isStarted(id),
         );
-        await this.writeUserConfig();
+        this.writeUserConfig();
```
KiranNiranjan, line 235: We don't need to await here?

KiranNiranjan: Got it using writeFileSync

KiranNiranjan (approved):

PR #1952, opened 2023-09-11
SDA-4288_main: Add config to allow / disallow user to edit podconfig

NguyenTranHoangSym:
## Description
Add config to allow / disallow user to edit podconfig

User can add this property via:
in Command line: msiexec /i Symphony.msi IS_POD_URL_EDITABLE="false"
in Config file: isPodUrlEditable


KiranNiranjan (approved):
If `isPodUrlEditable` is disabled we need to read from `globalConfig` file. 
Need to check if this logic is required.

PR #1975, opened 2023-10-10
SDA-4214: Update Toast notification to use toolbar

NguyenTranHoangSym:
## Description
Normal windows seem to work unexpectedly on Toast Notification
By using toolbar we can keep the notification config remained intact

### DIFF
--- a/src/renderer/notification.ts
+++ b/src/renderer/notification.ts
@@ -655,6 +655,7 @@ class Notification extends NotificationHandler {
       frame: false,
       transparent: true,
       fullscreenable: false,
+      type: 'toolbar',
       acceptFirstMouse: true,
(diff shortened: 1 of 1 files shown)


src/renderer/notification.ts
```diff
@@ -655,6 +655,7 @@ class Notification extends NotificationHandler {
       frame: false,
       transparent: true,
       fullscreenable: false,
+      type: 'toolbar',
```
KiranNiranjan, line 658: `toolbar` is only supported on Windows. Does this work on MacOS too?

KiranNiranjan (approved):

src/renderer/notification.ts
NguyenTranHoangSym: we have already handled it below, as 'Panel'

PR #1992, opened 2023-10-24
SDA-4370_23.11: Back port for SDA-4370 only

NguyenTranHoangSym:
## Description
Backport of https://github.com/finos/SymphonyElectron/pull/1988
but instead it will only has 
- Adding fix to the issue related to document.location.href changed will make pod url went unchanged
- Copy to clipboard will remove updatedHostName


KiranNiranjan:
No changes?

PR #2008, opened 2023-11-09
GitHub - use python version 3.11

KiranNiranjan:
GitHub actions recently upgraded the Python version on macOS runner https://github.com/actions/runner-images/pull/8695/files#diff-de83d5c5fa69158ec0c3eb7781d321cff841a277a7b393694b3cdbda34df0c9aR30

`distutils` is not present in Python version `3.12` which is required for `ffi-napi`


KiranNiranjan:
Closing this PR as the issue is fixed by removing an unused module which was causing the issue.

PR #2025, opened 2023-12-04
SDA-4077 (Add support for persisting client logs after reload)

KiranNiranjan:
## Description
- Add support for persisting client logs after reload
- Optimize writing client logs in chunks and in background

## Related PRs
https://github.com/SymphonyOSF/SFE-Lite/pull/16090

### DIFF
--- a/src/app/reports-handler.ts
+++ b/src/app/reports-handler.ts
@@ -88,7 +88,7 @@ const logTypes: string[] = [];
 const receivedLogs: ILogs[] = [];
 
 const validateFilename = (filename: string): string => {
-  return filename?.replace(/[^a-zA-Z0-9_.-]/g, '_');
+  return filename?.replace(/[^a-zA-Z0-9/_\.-]/g, '_');
 };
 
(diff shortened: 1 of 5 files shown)


src/app/reports-handler.ts
```diff
@@ -88,6 +87,23 @@ let logWebContents: WebContents;
...
 const receivedLogs: ILogs[] = [];
 
+const validateFilename = (filename: string): string => {
+  return filename?.replace(/[^a-zA-Z0-9_.-]/g, '_');
+};
+
+const writeClientLogs = async (retrievedLogs: ILogs[]) => {
+  for await (const logs of retrievedLogs) {
+    for (const logFile of logs.logFiles) {
+      const sanitizedFilename = validateFilename(logFile.filename);
```
KiranNiranjan, line 97: @sbenmoussati sanitized filename but `semgrep` is still failing here any idea?

KiranNiranjan: Yes, but that din't work either. so added `// nosemgrep`

PR #2045, opened 2024-01-02
SDA-4433: Clicking on the Close button in edit pod URL screen does not close the About

NguyenTranHoangSym:
## Description
Switch event from listening to Click into onMouseDown to cancel out blur from Input which is the main cause of these 2 problems:
- Click on the edge will instead return to normal aboutapp rather than edit aboutapp
- Click on **close** button will show flashing normal about app rather than close at edit aboutapp


spec/__snapshots__/aboutApp.spec.ts.snap
```diff
@@ -62,7 +62,7 @@ exports[`about app should render correctly 1`] = `
     <p
       className="AboutApp-copyright-text"
     >
-      Copyright © 2023 Symphony
+      Copyright © 2024 Symphony
```
KiranNiranjan, line 65: 🎉

KiranNiranjan (approved):

PR #2050, opened 2024-01-03
SDA-4427: Add fix to case of close and capture with sym hidden

NguyenTranHoangSym:
## Description
- Close will automatically shift focus back to main
- Upload will automatically shift focus back to main regardless to hidden or include SDA

### DIFF
--- a/src/app/stores/window-store.ts
+++ b/src/app/stores/window-store.ts
@@ -77,15 +77,17 @@ export class WindowStore {
     if (hideOnCapture) {
       this.restoreNotificationProperties();
       const storedWindows = this.getWindowStore();
-      let currentWindow = storedWindows.windows.find(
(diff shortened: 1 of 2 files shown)


src/app/stores/window-store.ts
```diff
@@ -177,6 +179,7 @@ export class WindowStore {
       if (windowToFocus) {
         windowToFocus?.show();
         windowToFocus?.moveTop();
+        // windowToFocus?.webContents?.focus();
```
KiranNiranjan, line 182: Remove this?

KiranNiranjan (approved):

src/app/stores/window-store.ts
NguyenTranHoangSym: done!

PR #2062, opened 2024-01-05
RTC-14731 - update "showScreenSharingIndicator" method

axeleriksson147:
`showScreenSharingIndicator` is a method used by Symphony Meetings. However, when we switched from Client 1.5 to 2.0, the window object for SymphonyElectron changed name and we haven't used this method in about 3 years. However, I am investigating https://perzoinc.atlassian.net/browse/RTC-14615 about screen sharing issues and one possible explanation is that our current way of sharing screen by posting messages via `window.postMessage` disappears or don't get picked up. Therefore we should look into using the exposed SymphonyElectron methods via the window object. It could be more reliable.

### DIFF
--- a/src/renderer/ssf-api.ts
+++ b/src/renderer/ssf-api.ts
@@ -512,12 +512,9 @@ export class SSFApi {
   }
 
   /**
-   * Shows a banner that informs user that the screen is being shared.
+   * Shows a banner that informs the user that the screen is being shared.
    *
-   * @param options object with following fields:
-   *    - stream https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/MediaStream object.
-   *             The indicator automatically destroys itself when stream becomes inactive (see MediaStream.active).
-   *    - displayId id of the display that is being shared or that contains the shared app
+   * @param options
    * @param callback callback function that will be called to handle events.
    * Callback receives event object { type: string }. Types:
    *    - 'error' - error occured. Event object contains 'reason' field.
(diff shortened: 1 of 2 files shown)


src/renderer/ssf-api.ts
```diff
@@ -512,12 +512,9 @@ export class SSFApi {
   }
 
   /**
-   * Shows a banner that informs user that the screen is being shared.
+   * Shows a banner that informs the user that the screen is being shared.
```
axeleriksson147, line 515: This method is now pretty much identical to `openScreenSharingIndicator`. Should we remove that one? Also I see `handleMessage` in `app-bridge.ts` says it's only for 1.5. Should that entire section of code perhaps be deleted? I suspect it's being used right now by the screen sharing but if we are making these changes I don't think anyone uses it then? @KiranNiranjan

src/renderer/ssf-api.ts
```diff
@@ -527,31 +524,25 @@ export class SSFApi {
...
       callback({ type: 'error', reason: 'bad displayId' });
       return;
     }
 
-    const destroy = () => {
-      throttledCloseScreenShareIndicator(stream.id);
-      stream.removeEventListener('inactive', destroy);
-    };
-
-    stream.addEventListener('inactive', destroy);
```
axeleriksson147, line 546: https://www.electronjs.org/docs/latest/api/context-bridge#parameter--error--return-type-support - This doesn't work due to these reasons. SFE-RTC will have to be responsible for closing the screen sharing indicator (which is probably better, SymphonyElectron shouldn't do too much).

src/renderer/ssf-api.ts
KiranNiranjan: Only RTC and client 1.5 was using the app-bridge.ts, now I think we will need remove the complete code

KiranNiranjan (approved):

PR #2074, opened 2024-01-11
RTC-14795 Abort launching of c9shell if network is offline

swasunb:


src/app/c9-shell-handler.ts
```diff
@@ -162,6 +163,13 @@ class C9ShellHandler {
    * Launches the correct c9shell process
    */
   private async _launchC9Shell(): Promise<ChildProcess | undefined> {
+    if (!windowHandler.isOnline) {
+      logger.info(
+        'c9-shell: launching of shell aborted due to network unavailability.',
```
alexblatov, line 168: @swasunb will we try again to launch C9 when we finally get the network?

alexblatov (approved):

src/app/c9-shell-handler.ts
swasunb: No, we need to manually start

PR #2077, opened 2024-01-15
SDA-4452 Focus on current window after taking a screenshot

sbenmoussati:
## Description
Mirror #2076


KiranNiranjan (approved):

src/app/screen-snippet-handler.ts
```diff
@@ -381,7 +385,6 @@ class ScreenSnippet {
           `screen-snippet-handler: close window failed with error: ${error}!`,
         );
       }
-
       webContents?.focus();
```
NguyenTranHoangSym, line 388: Will this be applied to close case while close is also has this issue?
I bet we also need focused window here

PR #2113, opened 2024-03-14
RTC-14942 Fix toast notifiaction vertical position

baphony:
https://perzoinc.atlassian.net/browse/RTC-14942

Notification position is fine but Toast notifications were vertically inverted on Windows:
image


swasunb (approved):

axeleriksson147 (approved):

KiranNiranjan (changes requested):
@baphony This is not a bug we intentionally positioned the **call** toast notification position exactly opposite to the **message** toast notification

baphony:
Thanks for the information, would it be good to have some information about it in the notification settings, close by the "Set position" button, to avoid confusion?

I close this PR

PR #2125, opened 2024-04-09
SDA-4528: Minimize Windows wont disappear on capture with SDA Hidden

NguyenTranHoangSym:
## Description
Fixing a bug relating to capture screenshot with application hidden,
If you minimize a popout and capture with app hidden, then the minimized popout will be temporarily disappeared until the next reload

### DIFF
--- a/src/app/stores/window-store.ts
+++ b/src/app/stores/window-store.ts
@@ -95,16 +95,19 @@ export class WindowStore {
       const fullscreenedWindows: IWindowState[] = [];
       // Restoring all windows except focused one
       storedWindows.windows.forEach((currentWindow) => {
-        if (currentWindow && currentWindow.isVisible) {
+        if (currentWindow) {
           const window: ICustomBrowserWindow | undefined = getWindowByName(
             currentWindow.id || '',
           ) as ICustomBrowserWindow;
           if (window) {
             if (currentWindow.isFullScreen) {
               fullscreenedWindows.push(currentWindow);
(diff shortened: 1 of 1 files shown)


src/app/stores/window-store.ts
```diff
@@ -95,16 +95,19 @@ export class WindowStore {
...
           if (window) {
             if (currentWindow.isFullScreen) {
               fullscreenedWindows.push(currentWindow);
-            } else if (!currentWindow.minimized && !currentWindow.focused) {
+            } else if (!currentWindow.focused) {
               window.showInactive();
               window.moveTop();
+              if (currentWindow.minimized) {
+                window.minimize();
+              }
```
KiranNiranjan, line 110: `currentWindow === window` here right? and do we need to call minimize again?

NguyenTranHoangSym: in the previous logic, we did not restore the minimized windows, that means after snippet done on execution, none of minimized windows get to be restored.
All I have to do is to trigger, showInactive on minimized but since showInactive will automatically restore minimized state of those Windows, thats why I will have to minimize it manually once again @KiranNiranjan

KiranNiranjan (approved):

src/app/stores/window-store.ts
NguyenTranHoangSym: @sbenmoussati as in my memory that capturing SDA hidden that means to completely hide away SDA without showing anything. So, thats the reason why hide was triggered on main and hide away all sub instances

PR #2270, opened 2025-01-23
Adding support for Windows to use navigator.mediaDevices.getDisplayMedia

axeleriksson147:
https://perzoinc.atlassian.net/browse/RTC-15321

https://github.com/finos/SymphonyElectron/pull/2266 - Builds on top of this PR. Adds a handler for systems (Windows, Mac < 15, Linux), that doesn't support `useSystemPicker`.


src/renderer/desktop-capturer.ts
```diff
@@ -139,16 +139,9 @@ export const getSource = async (
     }
   }
 
-  const updatedSources = sources
-    .filter((source) => source.name !== NOTIFICATION_WINDOW_TITLE)
-    .map((source) => {
-      return {
-        ...source,
-        ...{
-          thumbnail: source.thumbnail.toDataURL(),
```
axeleriksson147, line 148: Why can't `thumbnail` be the interface implementation instead of overriding it? I think it's just unnecessary.

alexblatov (approved):

src/app/display-media-request-handler.ts
```diff
@@ -1,19 +1,103 @@
...
+ * But for other versions and OSes, the code is executed.
  */
 export const setDisplayMediaRequestHandler = () => {
   const { defaultSession } = session;
-
   defaultSession.setDisplayMediaRequestHandler(
-    async (_request, _callback) => {
-      // TODO - Add support for Windows.
+    async (_request, callback) => {
+      logger.info('display-media-request-handler: getting sources');
```
KiranNiranjan, line 21: I think `isScreenShareEnabled` check should be added here.

axeleriksson147: `isScreenShareEnabled` is configured in Symphony.config in SDA, which allows/disallows screen sharing. Should this configuration really live in SDA? To me this is controlled via the entitlements in SFE-RTC? @swasunb what do you think?

My suggestion is that we add a check for system permission, but this extra media configuration could perhaps be deprecated here.

swasunb: Less reduce multiple sources of truth if we can

swasunb: what happens when there is no "Control access to screen" permission given to SDA? 
Do we have a posibility to ask it when you click on screenshare button? or nothing happens then?

axeleriksson147: As I understand it, on Windows there's no concept of screen share permission so it'll always be allowed. See this documentation - https://www.electronjs.org/docs/latest/api/system-preferences#systempreferencesgetmediaaccessstatusmediatype-windows-macos

Using the new `useSystemPicker` there seems to be a bug on Mac where even if I disallow the screen sharing still works.

Therefore, I think we're good to proceed with this.

KiranNiranjan (approved):

PR #2302, opened 2025-02-18
SDA-4732: Revamp the call notification

NguyenTranHoangSym:
## Description
Adding all new notification design changes.
the change will include
- Changes on IM-IM, Room - IM
- Changes on Federation Call, Unknown Federation Call
applying for both Dark / Light Mode


src/common/api-interface.ts
```diff
@@ -329,7 +333,14 @@ export interface ICallNotificationData {
   shouldDisplayBadge: boolean;
   acceptButtonText: string;
   rejectButtonText: string;
+  isFederatedEnabled?: boolean;
+  isPhone?: boolean;
+  notificationType?: string;
+  isPrimaryTextOverflowing?: boolean;
+  isSecondaryTextOverflowing?: boolean;
```
KiranNiranjan, line 340: Is this needed here? initially, this is just a local component state

src/renderer/components/call-notification.tsx
```diff
@@ -37,6 +38,10 @@ interface ICallNotificationState {
   isSecondaryTextOverflowing: boolean;
   isFederatedEnabled: boolean;
   zoomFactor: number;
+  isPhone?: boolean;
+  notificationType?: string;
+  callerNumbers?: string;
```
KiranNiranjan, line 43:
```suggestion
  callerNumber?: string;
```

src/renderer/components/call-notification.tsx
```diff
@@ -365,11 +432,24 @@ export default class CallNotification extends React.Component<
         callType === 'IM'
           ? 'profilePlaceHolderContainer'
           : 'roomPlaceHolderContainer';
+
+      if (!callerName) {
```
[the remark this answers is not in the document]
NguyenTranHoangSym: This will exclusively apply onto Fed. Normally, if a person calls to you without having any identification (saving contact) it will now show the User Icon instead of name character.
IF this is not a fed user this will return empty string, this section wont show either

src/common/api-interface.ts
NguyenTranHoangSym: Thats correct. Might be confused by the model

KiranNiranjan (approved):

KiranNiranjan (approved):

PR #2344, opened 2025-03-19
RTC-15423 - Remove old screen share code

axeleriksson147:
https://perzoinc.atlassian.net/browse/RTC-15423

The `getMediaSource` API was replaced with the code in `display-media-request-handler.ts`. This PR removes all usage of it.


src/app/child-window-handler.ts
axeleriksson147, line 1: In this file I believe `openScreenPickerWindow` and `createScreenPickerWindow` should be removed as well. In `display-media-request-handler.ts`, we create it in the following way:

`this.screenPickerWindow = createComponentWindow(
          'screen-picker',
          browserWindowOptions,
        );
`

src/renderer/app-bridge.ts
```diff
@@ -12,11 +12,6 @@ import {
   IScreenSnippet,
```
KiranNiranjan, line 12: @sbenmoussati  I think we could completely delete this file

KiranNiranjan (approved):

src/app/child-window-handler.ts
axeleriksson147: Maybe we can do it in a subsequent PR @KiranNiranjan ? I think the scope of this one is sufficient.

src/app/child-window-handler.ts
```diff
@@ -338,13 +338,6 @@ export const handleChildWindow = (webContents: WebContents): void => {
         if (browserWin.webContents) {
           // validate link and create a child window or open in browser
           handleChildWindow(browserWin.webContents);
-
-          // Updates media permissions for preload context
-          const { permissions } = config.getConfigFields(['permissions']);
```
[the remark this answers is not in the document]
axeleriksson147: We have the `canProjectScreenShare` entitlement in SFE-RTC which will disallow the usage if set to false. This permission check is set in `Symphony.config`. I think it's unnecessary to have a check in both `SFE-RTC` and `SymphonyElectron`. @swasunb agree?

PR #2361, opened 2025-04-24
RTC-15423 - Remove unused screen share methods

axeleriksson147:
https://perzoinc.atlassian.net/browse/RTC-15423

I think these two methods are left to clean up the old way of doing screen share.


axeleriksson147:
😢

PR #2383, opened 2025-05-28
SDA-4909_25.5: Neutralize Path if they arent exist

NguyenTranHoangSym:
## Description
Neutralize file before removing


KiranNiranjan:
@NguyenTranHoangSym, I think the issue is coming from https://github.com/finos/SymphonyElectron/pull/2383/files#diff-95038264eb9cb3f0bcd0a77c0e0d7905fd9bc041d35ec3bf0dfc4d15c3b651c7R327 

Move the ` try catch` block within the loop so that one failure does not block exporting the rest of the files.

KiranNiranjan (approved):

PR #2439, opened 2025-07-09
SDA-4937_4944: Adding border radius to window 11 and handle bleeding edges

NguyenTranHoangSym:
## Description
- Remove bleeding edges in all windows while zooming
- Support border radius in new acrylic / mica design of Windows
Applied only for Notifications (Call / Toast) and Screen Sharing

### DIFF
--- a/src/app/window-utils.ts
+++ b/src/app/window-utils.ts
@@ -21,7 +21,13 @@ import { format, parse } from 'url';
 import { productDisplayName } from '../../package.json';
 import { apiName, EPresenceStatusGroup } from '../common/api-interface';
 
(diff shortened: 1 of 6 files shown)


src/app/window-utils.ts
```diff
@@ -211,6 +211,8 @@ export const createComponentWindow = (
     show: false,
     width: 300,
     ...opts,
+    roundedCorners: isWindowsOS ? true : false,
```
KiranNiranjan, line 214: Component windows are not affected right? Do we need modify this here?

NguyenTranHoangSym: roundedCorner will apply to every frame. Because right now those frames are also sharped edge

KiranNiranjan (approved):

